Issue #33, opened 2023-10-13
Better user feedbacks and errors in app


DivadNojnarg:

DivadNojnarg:
Closing an will open a new issue when it makes sense to tackle this again.

PR #56, opened 2023-10-26
Rework computation authorization between blocks

### LINKED ISSUE #55: Reactivity inconsistencies in generate_server.transform_block
```r
o <- observeEvent(
        eval(obs_expr(blk())),
        secure(eval(set_expr(blk()))),
        ignoreInit = TRUE
      )
```

This gets triggered too many times. Hashing inputs shows that the third event should not trigger. 

```bash
[1] "35f073918d0c815bbf81af029fa4a3f8"
Called from: observe()
Browse[1]> c
[1] "e36cc6a7ecb1cc806c044ad143937f7a"
Called from: observe()
Browse[1]> c
[1] "e36cc6a7ecb1cc806c044ad143937f7a"
Called from: observe()
Browse[1]>
```

Also, I don't think we should allow this expression to run without checking that the block input state is compatible with the received data. For instance, assuming input values refer to columns not available in the `in_dat()`...

```r
out_dat <- reactive({
   evaluate_block(blk(), data = in_dat())
})
```

DivadNojnarg:
Will fix #55 and #33.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -205,7 +205,7 @@ initialize_block.data_block <- function(x, ...) {
 new_filter_block <- function(
     data,
     columns = colnames(data)[1L],
-    values = character(),
+    values = data[1, colnames(data)[1L]],
     filter_fun = "==",
     ...) {
   sub_fields <- function(data, columns) {
--- a/R/server.R
+++ b/R/server.R
@@ -179,7 +179,7 @@ generate_server.transform_block <- function(x, in_dat, id, ...) {
       # the validator is FALSE. Avoids to
       # send wrong data to the next block.
       out_dat <- reactive({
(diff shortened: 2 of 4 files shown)


DivadNojnarg:
@JohnCoene : for now validation is not added to the data and plot block (only transform blocks). If you want to have it in all blocks, let me know.

JohnCoene:
Currently I get `Error in module: object 'InputValidator' not found`

DivadNojnarg:
When the block is invalid and collapsed, we add an extra border class to signal that something is wrong in this block.
Screenshot 2023-10-27 at 10 34 43

DivadNojnarg:
Closing as implemented in another PR.

PR #60, opened 2023-10-27
Create new block

JohnCoene:
This adds `create_block` to create `plot` and `transform` blocks skeletons


DivadNojnarg:
You can fix the lint (as much as you can) and merge

JohnCoene:
Do we run styler on this? Which style?

DivadNojnarg:
We have a .lintr file and we run lintr::lint_package (https://github.com/blockr-org/blockr/blob/main/.github/workflows/lint.yml). I used styler at the very beginning with the default settings (just styler::style_*).

PR #70, opened 2023-11-14
mutate block

christophsax:
It works now, also with several expressions. Still buggy when I add and remove rows, but this could be probably addressed by looking at the `ace.R` only.

New classes `mutate_block`, `field_namedchar`. 

TBD in the afternoon, @nbenn @DivadNojnarg


R/ui.R
```diff
@@ -119,10 +119,11 @@ generate_ui.stack <- function(x, id = NULL, ...) {
         })
       )
     ),
-    sortable::sortable_js(
```
christophsax, line 122: @DivadNojnarg interferes with shinyAce

JohnCoene: What problem does it cause? Perhaps the ace editor uses the `block` class.

DivadNojnarg: Sortable is also preventing from being able to copy and paste within a block, and so far we don't use it because we don't support rearranging blocks. 

https://github.com/blockr-org/blockr/assets/18291543/d6fefb2d-7725-463d-aab3-a87febad82cd

JohnCoene: We'll have to use a handle for the sortable I guess.

JohnCoene: Fix is here https://github.com/blockr-org/blockr/pull/71

R/mutate-block.R
```diff
@@ -70,10 +68,10 @@ generate_server.mutate_block <- function(x, in_dat, id, ...) {
            blk_updated <- update_fields(
              r_blk(), session,
              in_dat(),
-             value = r_value(),
-             expression = mutate_expr(r_value())
+             value = r_value()
            )
 
+           attr(blk_updated,"expr") <- mutate_expr(r_value())
```
christophsax, line 74: @nbenn not sure how canonical, but this saves us from the hidden field

christophsax:
@DivadNojnarg Do we need the code coverage checks? It takes a lot of time to keep the coverage, and the benefit seems minimal to me?

R/ui.R
christophsax, line 1: @DivadNojnarg sorry, I merged this too quickly, will be more careful in the future. This was needlessly invasive and perhaps the reason for John's issue. Please let me know if I should revert.

PR #91, opened 2023-11-21
fix: collapse logic closes #90

### LINKED ISSUE #90: Collpase/toggle of items erroneous
Collapse of blocks and elements within not working properly

JohnCoene:
Fixes #90 

Collapse logic of blocks is not consistent


DivadNojnarg (changes requested):
Thanks for looking into this. The initial state looks weird (see video), as I am unable to see the second block toolbar to collapse/uncollapse the block. 

https://github.com/blockr-org/blockr/assets/18291543/e76fe220-0d7f-4680-81e0-593ae29b533f

JohnCoene:
The new validation breaks much of the logic of how we display blocks, namely with the filter block.

DivadNojnarg:
I don't think this is related, the test was recorded from your branch, which does not contain the new validation system. Below another test without the filter on the same branch and cleared browser cache.


https://github.com/blockr-org/blockr/assets/18291543/37bdcd74-6407-4c3b-aa91-f1fe1ca38560

JohnCoene:
That was caused by:

1. change in https://github.com/blockr-org/blockr/blob/5a3afa85e081f0783db341db51d96fa75ed966d9/R/ui.R#L17
2. Blocks like filter not having a last output

Solution:
1. Revert block UI change so it defaults to hidden
2. Add loading state

JohnCoene:
Ready to merge

DivadNojnarg:
Thanks @JohnCoene! The collapsing works well now. However I still see a spinner in the data block.


https://github.com/blockr-org/blockr/assets/18291543/ef65a7f5-6d6c-432e-a620-587854f3fddb

JohnCoene:
Sorry @DivadNojnarg thanks for catching that.

PR #97, opened 2023-11-27
Block regsitry

nbenn:


nbenn:
Blocks can now be registered with blockr by calling `register_block()`. This may happen in an `.onLoad()` hook as is the case for the test pkg

https://github.com/blockr-org/blockr/blob/82ede9ccbb53f484ad47183fab3e703479fc1d0f/inst/testdata/test.registry/R/zzz.R#L1-L9

or in a script together with the constructor definition (and potential `registerS3method()` calls).

In order to list all registered blocks we have `available_blocks()` and getters `block_name()`/`block_descr()`. If more is needed here please let me know @JohnCoene.

nbenn:
@DivadNojnarg I need your feedback here to be able to merge (as per the new merging rules).

R/registry.R
```diff
@@ -0,0 +1,148 @@
...
+}
+
+register_blockr_blocks <- function(pkg) {
+
+  if (missing(pkg)) {
+    pkg <- pkg_name()
+  }
+
+  register_blocks(
+    c(new_data_block, new_filter_block, new_select_block, new_summarize_block),
```
DivadNojnarg, line 134: I would explicitly pass the parameter names. A new dev would not necessarily remember about what `register_blocks` takes as input.

R/registry.R
```diff
@@ -0,0 +1,148 @@
...
+      pkgs <- pkgs[ids]
+    }
+
+    ids <- names(pkgs)[lgl_ply(pkgs, identical, package)]
+  }
+
+  rm(list = ids, envir = block_registry, inherits = FALSE)
+}
+
+register_blockr_blocks <- function(pkg) {
```
DivadNojnarg, line 127: why `register_blockr_blocks` and not simply `register_blocks`

DivadNojnarg: or maybe because `register_blockr_blocks` is for internal use and `register_blocks` for external use. If yes, ignore my comment

inst/testdata/test.registry/R/zzz.R
```diff
@@ -0,0 +1,13 @@
+.onLoad <- function(libname, pkgname) {
+
+  blockr::register_block(
+    new_head_block, "head block", "return first n rows",
```
DivadNojnarg, line 4: same comment: adding parameter names would be more dev/user friendly.

R/registry.R
```diff
@@ -0,0 +1,148 @@
...
+
+#' @param constructor Block constructor
+#' @param name,description Metadata describing the block
+#' @param classes Block classes
+#' @param input,output Object types the block consumes and produces
+#' @param package Package where block is defined
+#'
+#' @rdname available_blocks
+#' @export
+register_block <- function(constructor, name, description, classes, input,
```
DivadNojnarg, line 70: ctor, constructor? Should we use ctor instead as it is used everywhere else in the code?

DivadNojnarg (commented):
Thanks @nbenn, this looks good for a first iteration. I tested locally, it works well. Only minor comments regarding a parameter name (ctor vs constructor) and explicitly naming params when calling register_blocks. I got also a bit confused between register_blockr_blocks (internal for blockr blocks) and register_blocks (external API).

JohnCoene:
@nbenn Can we add the function name?

The information in the registry is used UI-side to communicate to the server what block has to be added and where, before I use to pass the name of the block from the UI to the server, I don't think we have something equivalent here.

```r
r$> blockr::block_name()                                                        
[1] "data block"      "filter block"    "select block"    "summarize block"

r$> blockr::block_descr()                                                       
[1] "choose a dataset"          "filter rows in a table"   
[3] "select columns in a table" "summarize data groups"
``` 

Only thing I think of is to `gsub` the spaces in the name with `_` and assume that's the name of the corresponding function.

R/registry.R
nbenn: Sure, never bad to be explicit.

R/registry.R
nbenn: Yes, that is the case.

inst/testdata/test.registry/R/zzz.R
nbenn: Yup.

R/registry.R
nbenn: I opted for the other way round to stick to `constructor`. A bit more verbose, but I did not want the user facing argument to be called `ctor`, as the meaning then might be a bit less obvious.

DivadNojnarg (approved):
Good to merge. Thanks.

nbenn:
@JohnCoene I'm afraid the function name won't be a robust way to identify the constructor, as you'd then also need to know the environment where this name is bound. So far this has been the "package" environment (the environment of exported functions) of blockr. But this is not the case for 3rd party block implementations.

I am holding on to the package name of the "source" of a block constructor. But more for convenience when unloading things.

A third party block constructor might also be un-exported. Or not be defined in a package at all.

For that reason I bind the constructor also in my "registry" environment, so that I have a unified place to look up these functions. I added a convenience function `construct_block()` that helps you with this[^1]. You can pass the name of the block (or a block "description" object to that and it will call the constructor for you (with `...` forwarded to the constructor if that is relevant).

To expose these names to the user, the getters `block_name()` and `block_descr()` return named character vectors and what you get from `available_blocks()` is a named list. These names can be passed to `construct_block()`.

[^1]: The function name `create_block` (instead of `construct_block`) was my first choice, but this was already taken. Do we want to reconsider here?

JohnCoene:
@nbenn sorry I missed `construct_block`, that'll work just fine!

nbenn:
@JohnCoene all good, you did not miss anything. I just added that 1h ago.

Issue #79, opened 2023-11-18
Generate Code Method


JohnCoene:
Should things like these be methods?

https://github.com/blockr-org/blockr/blob/9c88b24a7597ff86af5b69714f25952778d391da/R/block.R#L78

Extending for new block may require a special case/`if` and a PR in {blockr}?

christophsax:
It could be done in a better way. Use S3 dispatch.

DivadNojnarg:
Closed in #107.

PR #107, opened 2023-11-27
will fix Generate Code Method #79

DivadNojnarg:
Cleanup dirty code chunk.


nbenn (commented):
The change is an improvement for sure. What I don't understand here is why the `type` setting per field

https://github.com/blockr-org/blockr/blob/8f9a24a26bfd0cc3fa143fa893cf76c657c0d4ed/R/field.R#L14-L15

cannot be used for this. This feels like the canonical piece of infrastructure to handle this. Do we somehow need to make that more capable to cover additional use-cases?

R/utils.R
```diff
@@ -184,7 +184,7 @@ dropNulls <- function(x) {
 #'
 #' @keywords internal
 convert_block <- function(from = new_select_block, to, data, ...) {
-  block <- initialize_block(from(data, ...), data)
+  block <- initialize_block(from(data, type = "name", ...), data)
```
DivadNojnarg, line 187: @nbenn: I suppose this is what you meant about field type.

R/utils.R
```diff
@@ -137,7 +137,7 @@ type_trans <- function(x) {
 
   switch(attr(x, "type"),
     literal = res,
-    name = as.name(res)
+    name = unlst(lapply(res, as.name))
```
DivadNojnarg, line 140: type_tans was not able to deal with values of length > 1 (only returned the first element). I made this change here to ensure as.name applies to any element of res.

R/utils.R
```diff
@@ -184,8 +184,10 @@ dropNulls <- function(x) {
 #'
 #' @keywords internal
 convert_block <- function(from = new_select_block, to, data, ...) {
-  block <- initialize_block(from(data, type = "name", ...), data)
+  block <- initialize_block(from(data, ...), data)
   class(block)[[1]] <- sprintf("%s_block", deparse(substitute(to)))
+  # Change type to name since arrange/group_by don't work with literals
+  attr(block$columns, "type") <- "name"
```
DivadNojnarg, line 190: changed field type from literal to name so that `type_trans` can do its work with arrange and group_by blocks.

nbenn: This has nothing to the with the PR, so feel free to ignore for now.

I really think we should get rid of `convert_block()`. If several blocks share functionality that we do not want to duplicate, we can always abstract this into functions and use those in multiple places.

R/utils.R
nbenn: Generally I feel that wrapping things with `unlist()` to make things work leads to other problems.

We might want to handle more explicitly the case where `value(x)` for a field is vector-valued.

Where does this cause issues? For list fields? I assume a range field would also not have a scalar `value()`, but you'd only want to use that as a literal in an expression.

I'm fine with just merging and sorting out this separately though.

nbenn (commented):
I'm fine with merging. We can sort out the issues separately.

R/utils.R
DivadNojnarg: Agree. When we split block into core and extension, we can rework this.

DivadNojnarg:
@nbenn: it seems I need re-approval. It is still blocked.

nbenn (commented):
Lgtm.

nbenn (approved):
All good.

Issue #92, opened 2023-11-21
Workspace class


christophsax:
As discussed today, it would make sense to prioritize it, despite what we said last Friday. @nbenn is drafting, and discusses with @JohnCoene afterwards.

christophsax:
We schedule a meeting next week, Tue 5pm. A lot to discuss and think about.

christophsax:
Shiny logic in #121

DivadNojnarg:
After #124, it is possible that the workspace has to be a structure, similar to what we have for stacks and blocks.

PR #124, opened 2023-12-08
First pass into workspace UI/server

DivadNojnarg:
TO DO:

- add tests.
- <del>add way to edit stacks: pass blocks</del>. 
- <del>Important: Cleanup stack inputs? observers and output (like for blocks). Below is how `input` looks like after stacks are added, cleared and reset </del> (OUT OF THE SCOPE OF THIS PR).

```r
<ReactiveValues> 
  Values:    myworkspace-add_stack, myworkspace-clear_stacks, myworkspace-mystack-1-jgnzeezgtmhotlp-dataset, myworkspace-mystack-1-jgnzeezgtmhotlp-remove, myworkspace-mystack-1-jgnzeezgtmhotlp-res_cell_clicked, myworkspace-mystack-1-jgnzeezgtmhotlp-res_cells_selected, myworkspace-mystack-1-jgnzeezgtmhotlp-res_columns_selected, myworkspace-mystack-1-jgnzeezgtmhotlp-res_rows_all, myworkspace-mystack-1-jgnzeezgtmhotlp-res_rows_current, myworkspace-mystack-1-jgnzeezgtmhotlp-res_rows_selected, myworkspace-mystack-1-jgnzeezgtmhotlp-res_search, myworkspace-mystack-1-jgnzeezgtmhotlp-res_state, myworkspace-mystack-1-last_changed, myworkspace-mystack-1-remove, myworkspace-mystack-1-title, myworkspace-mystack-2-last_changed, myworkspace-mystack-2-remove, myworkspace-mystack-2-title, myworkspace-mystack-2-wpmbppudytffvdu-dataset, myworkspace-mystack-2-wpmbppudytffvdu-remove, myworkspace-mystack-2-wpmbppudytffvdu-res_cell_clicked, myworkspace-mystack-2-wpmbppudytffvdu-res_cells_selected, myworkspace-mystack-2-wpmbppudytffvdu-res_columns_selected, myworkspace-mystack-2-wpmbppudytffvdu-res_rows_all, myworkspace-mystack-2-wpmbppudytffvdu-res_rows_current, myworkspace-mystack-2-wpmbppudytffvdu-res_rows_selected, myworkspace-mystack-2-wpmbppudytffvdu-res_search, myworkspace-mystack-2-wpmbppudytffvdu-res_state, myworkspace-mystack-3-last_changed, myworkspace-mystack-3-remove, myworkspace-mystack-3-title, myworkspace-mystack-3-uozygoqqaknabbd-dataset, myworkspace-mystack-3-uozygoqqaknabbd-remove, myworkspace-mystack-3-uozygoqqaknabbd-res_cell_clicked, myworkspace-mystack-3-uozygoqqaknabbd-res_cells_selected, myworkspace-mystack-3-uozygoqqaknabbd-res_columns_selected, myworkspace-mystack-3-uozygoqqaknabbd-res_rows_all, myworkspace-mystack-3-uozygoqqaknabbd-res_rows_current, myworkspace-mystack-3-uozygoqqaknabbd-res_rows_selected, myworkspace-mystack-3-uozygoqqaknabbd-res_search, myworkspace-mystack-3-uozygoqqaknabbd-res_state 
  Readonly:  TRUE 
```


R/ui.R
```diff
@@ -150,6 +150,41 @@ generate_ui.stack <- function(x, id = NULL, ...) {
...
+#' @rdname generate_ui
+#' @export
+generate_ui.workspace <- function(x, id = NULL, ...) {
+  stopifnot(...length() == 0L)
+
+  id <- if (is.null(id)) attr(x, "name") else id
+
+  ns <- NS(id)
+
+  stacks <- get_workspace_stacks()
```
DivadNojnarg, line 162: this feels actually a bit weird to have to get the stacks this way since workspace is not a structure like `new_stack` or `new_block`. I can make it work but this does not feel super in line with the rest of our API.

R/ui.R
```diff
@@ -150,6 +150,41 @@ generate_ui.stack <- function(x, id = NULL, ...) {
...
+        span(class = "badge bg-secondary", textOutput(ns("n_stacks")))
+      ),
+      actionButton(ns("clear_stacks"), "Clear all", icon("trash"), class = "bg-danger")
+    ),
+    div(
+      class = "workspace",
+      div(
+        class = "d-flex justify-content-between stacks",
+        lapply(seq_along(stacks), \(i) {
+          generate_ui(stacks[[i]], id = ns(sprintf("mystack-%s", i)))
```
DivadNojnarg, line 181: Same as above. I would have expected to loop over x (the workspace) and not over the result of `get_workspace_stacks`.

DivadNojnarg:
Note: #119 is needed to cleanup module server after being removed (cc @nbenn)

DivadNojnarg:
Still to do:
- maybe add the "Add block button" within the stack card.
- improve layout.

Screenshot 2024-01-10 at 14 55 48

DivadNojnarg:
Few notes for @nbenn:
- We'll need to handle the server conflicts with the main branch: from what we discussed, I propose to accept both incoming and current to be able to use the old approach for `generate_server`. 
- Now that we have the workspace, stacks, blocks, ... a good time to rethink the validation process.
- This is a basic/default workspace implementation. @JohnCoene has his own with the super framework.

R/ui.R
```diff
@@ -301,6 +367,48 @@ stack_header.stack <- function(x, title, ns, ...) {
...
+#' @rdname generate_ui
+#' @export
+generate_ui.workspace <- function(x, id = NULL, ...) {
+  stopifnot(...length() == 0L)
+
+  id <- if (is.null(id)) attr(x, "name") else id
+
+  ns <- NS(id)
+
+  stacks <- get_workspace_stacks()
```
nbenn, line 379: This is more a note-to-self: my workspace API here is not good. If the workspace is passed as `x` you'd want to be able to call `get_workspace_stacks()` as `get_workspace_stacks(x)`.

R/server.R
```diff
@@ -396,9 +396,169 @@ handle_remove_block.block <- function(x, vals, session = getDefaultReactiveDomai
...
+#' @rdname generate_server
+handle_remove.stack <- function(x, vals, session = getDefaultReactiveDomain(), ...) {
+  input <- session$input
+  id <- attr(x, "name")
+  observeEvent({
+    input[[sprintf("remove-stack-%s", id)]]
+  }, {
+    # We can't remove the data block if there are downstream consumers...
+    stacks <- get_workspace_stacks()
+    to_remove <- which(chr_ply(stacks, \(x) attr(x, "name")) == id)
```
nbenn, line 415: This is not wrong. From a design-perspective, it might be worthwhile to consider either passing the index from the parent scope, where it should be more easily available, no? This is also in anticipation of maybe not implementing the workspace as environment.

R/server.R
```diff
@@ -396,9 +396,169 @@ handle_remove_block.block <- function(x, vals, session = getDefaultReactiveDomai
...
+#' @export
+generate_server.workspace <- function(x, id = NULL, ...) {
+  stopifnot(...length() == 0L)
+
+  id <- if (is.null(id)) attr(x, "name") else id
+
+  moduleServer(
+    id = id,
+    function(input, output, session) {
+      vals <- reactiveValues(stacks = list(), new_blocks = list())
```
nbenn, line 438: More of a naming issue (probably my understanding here is incomplete), but why the plural? Does `new_blocks` actually hold on to multiple new blocks at the same time?

R/ui.R
```diff
@@ -206,6 +206,40 @@ generate_ui.block <- function(x, id, ..., .hidden = !getOption("BLOCKR_DEV", FAL
...
+      `aria-controls` = ns("addBlockCanvas")
+    ),
+    off_canvas(
+      id = ns("addBlockCanvas"),
+      title = "My blocks",
+      position = "bottom",
+      radioButtons(
+        ns("selected_block"),
+        "Choose a block",
+        choices = setNames(seq_len(length(list_blocks())), list_blocks()),
```
nbenn, line 235: We have `set_names()` in `utils.R`, b/c I thought it was unnecessary to import an entire package for a 2-line function. This effort was thwarted by @christophsax in https://github.com/blockr-org/blockr/commit/3e1b971120423ca9d251ad44933f3dc9fdeef730.

nbenn (approved):
Lgtm. We should tweak the workspace API a bit. If in your scope you have the workspace available as `x`, it feels wrong to call e.g. `get_workspace_stacks()` instead of `get_workspace_stacks(x)`. Happy to discuss your thoughts on whether the workspace object should remain implemented as environment or not.

R/server.R
DivadNojnarg: Yes that's TRUE. This is a wrong naming because to the best of my knowledge, we only add 1 block at a time. Can you confirm @JohnCoene? That should also apply to the stack server as well:

```r
#' @param new_blocks For dynamically inserted blocks.
#' @export
generate_server.stack <- function(x, id = NULL, new_blocks = NULL, ...) {
...
}
```

I can change it.

R/ui.R
DivadNojnarg: Another remark: after checking with @JohnCoene in another PR, I might be wrong using `list_blocks()` which is not exported. I should use `available_blocks()` instead.

R/ui.R
DivadNojnarg: Yes this is weird. We can leave it as is if the time constaint is too high with other priorities. Anyway we still have #92 open so we can add a bullet point here.

R/server.R
JohnCoene: Yes, this is correct.

I cannot run `generate_server` on the workspace directly. I need to regenerate their UI on the correct tabs and their servers in their correct tabs/modules as well as pass the adequate `new_block` reactive to each server.

PR #127, opened 2023-12-13
WIP redo names/IDs

nbenn:


DivadNojnarg:
@nbenn: I would also need the block ID refactoring to handle the remove block feature.

PR #129, opened 2023-12-14
JSON serialization

nbenn:
For now this is WIP, will add serialization for

- [x] fields
- [x] blocks
- [x] stacks
- [x] workspace


nbenn:
For json serialization we now have something that works (but is not super elegant). We can do

```r
stack <- new_stack(
  data_block,
  mutate_block
)

json <- to_json(stack)

writeLines(json, "stack.json")

# optionally restart the session here

json <- readLines("stack.json")

stack <- from_json(json)

options("BLOCKR_DEV" = TRUE)
serve_stack(stack)
```

Under the hood, i'm writing the code into the json that is necessary to re-generate the objects (something like `dput()`, but more sophisticated using the constructive package by @moodymudskipper). This suffers from the problem that functions of the re-created objects are not associated with the correct environments (along with an incorrect inheritance chain).

(With "functions" I mainly am referring to the situation where a function is passed as field argument.)

While this will have to be re-worked, as it will most certainly lead to issues (for example if functions behave like closures), it is meant as prototype to be able to use the basic functionality this feature should provide.

JohnCoene:
I think blockr should provide serialisable objects rather than the serialisation itself since that happens at the app level.

See example below.

```r
stack1 <- new_stack(data_block)
stack2 <- new_stack(data_block)

# serialise at stack level ~ to_json.stack
stacks <- list(
    to_json(stack1),
    to_json(stack2)
)

json <- jsonlite::toJSON(list(stacks = stacks, title = "App"))

jsonlite::fromJSON(json)
```

nbenn:
@JohnCoene I'm not sure I follow.

1.
    > I think blockr should provide serialisable objects
    
    Does it not already to exactly that? Cf.
    
    ```r
    stk_orig <- list(
      stack1 = new_stack(data_block),
      stack2 = new_stack(data_block)
    )

    json <- jsonlite::toJSON(
      list(stacks = lapply(stk_orig, to_json), title = "App")
    )

    stk_rest <- lapply(jsonlite::fromJSON(json)[["stacks"]], from_json)

    # do somethig with your restored stacks
    ```

2. > rather than the serialisation itself since that happens at the app level

    You can save yourself a bit of trouble by ser/deserializing directly the entire workspace, which I thought was the object we were using to encapsulate an apps list of stacks (+ metadata). The API here is quite simple
    
    ```r
    set_workspace(
      stack1 = new_stack(data_block),
      stack2 = new_stack(data_block),
      title = "App" 
    )

    json <- to_json()

    restore_workspace(json, force = TRUE)
    ```

JohnCoene:
I don't understand, why insist on wrapping this in `toJSON`?

https://github.com/blockr-org/blockr/blob/4f2f70e29f4ec695ba0e095685be84db1cf530a2/R/json.R#L20

We don't need JSON we need the list: 

https://github.com/blockr-org/blockr/blob/4f2f70e29f4ec695ba0e095685be84db1cf530a2/R/json.R#L21

I know it "works" but it's really not something that is advised: we serialise an object to a _JSON string_ that we place in another JSON or worse, another file type like putting JSON in a yaml?

If you don't want to change it it's fine by me but I'm indeed going to have to do a lot of things like this: `stk_rest <- lapply(jsonlite::fromJSON(json)[["stacks"]], from_json)` which is not very elegant.

nbenn:
@JohnCoene At some point we agreed that

1. we have a workspace object that contains stacks and downstream objects, as well as some metadata,
2. we would serialize/deserialize this workspace object.

that is what was trying to do here. The PR (started on Dec 14.) is titled "JSON serialization" after all. I assume in the mean time, for you at least, this has changed. Do you maybe want to share what it is that *you're* trying to achieve? It's hard for me to do anything meaningful if I'm left in the dark on the intended use.

We had this discussion on storing "settings" as part of the workspace object etc. I asked for your feedback, but instead of requesting changes and improving on what we have, you seem to just have abandoned the idea entirely.

I believe I now understand what it is that you want this PR to look like. And I'm fine with providing this (see update). But why don't we discuss such things a bit more? If we agree on using this workspace object in a certain way and with time you feel changes are needed, why don't you say so (and maybe show an example of a problem that you have with the given architecture) instead of just silently taking another direction?

cc @christophsax @DivadNojnarg

nbenn:
@JohnCoene just a word of caution when using the now provided & exported (lower level) `blockr_serialize()` output: the representation this provides is subject to change. Having a list with components `class`, `payload` and `blockr` is an (arbitrary) implementation detail and is not intended to be part of a stable API. What you should be able to rely on is safely round tripping `blockr_serialize()`/`blockr_deserialize()`. Does that make sense?

JohnCoene:
I'm going to write some thoughts here as I go through the implementation/use of this in an application. It's not definitive as I still have things to explore but it may help us move forward.

My first realisation is that here we actually don't directly work with stacks. I never truly assign a stack to the workspace because I have to work with the reactive representation of the stack.

https://github.com/blockr-org/blockr/blob/ee492769a901e2301b887d7a3074109cdf5ff4d2/R/server.R#L249

We have to react on changes in the stack to push it to the workspace and so we can ensure that when we save/serialise we capture the latest/current version of the stacks. 
This logic could be somewhat changed but it will remain that we will have to handle the reactives rather than `stack` itself. (the `stack` object will always be empty, blocks will only be present in it's reactive "copy" found in `generate_server`).

With that, I'm wondering if that should not be implemented here in blockr rather than outside of blockr.

If we want this managed outside of blockr, it would be great if we could add a method/function on the workspace so that we add/update a stack to the workspace, a sort of upsert for the workspace. This is because I generally don't need to assign a series of stacks to the workspace at once: we observe a change on a stack and we want to push/update it in the workspace (essentially a way `assign` to the workspace environment).

Note: we can get the reactive stack with

```r
server <- generate_server(stack)
# server$stack
```

For reasons I do not understand.

```r
# works
server <- generate_server(new_stack())
set_workspace(x = server$stack)

# doesn't work when adding a block dynamically
server <- generate_server(new_stack(), new_block = reactive(...))
set_workspace(x = server$stack)
```

This check [here](https://github.com/blockr-org/blockr/blob/ee492769a901e2301b887d7a3074109cdf5ff4d2/R/workspace.R#L49C22-L49C53) fails because for some reason after dynamically adding a block to stack `server$stack` loses its `stack` class.

I can make it "work" by forcing the `stack` class back on it (`class(x) <- "stack"`) but it's probably not advised and it may still leave things downstream broken (e.g.: restore) but that is yet to be tested.

I will work on this some more and may come back with more thoughts/feedback.

JohnCoene:
Could we slightly modify this object?

https://github.com/blockr-org/blockr/blob/ee492769a901e2301b887d7a3074109cdf5ff4d2/R/json.R#L64

It's a bit of a strange structure.

```r
names(desirialised_stacks)

#> [1] "sbkzzcyxomkcttm" "sqcmlzaqypfasiu" "nuarkrskwkdexwh" "title" "settings"
```

Could we put the stacks in their own list/object?

JohnCoene:
Similar to previous comment, we can delete a stack: could we have a function to remove a stack from the workspace?

DivadNojnarg:
@nbenn  and @JohnCoene. I gave it a quick try with the internal block workspace shiny implementation we have in main + I added a piece of UI/server to call serialise and restore workspace from shiny. Then I do:

- I start my demo with a workspace of 3 stacks each containing a data block.
- I remove the third stack.
- I add a select block to the first stack.
- I click on serialize. The value is saved in a reactiveValue.
- I click on restore which calls `blockr_deserialize(<SAVED_VALUE>)`.

The object I get contains 2 stacks but there is no select block in the first stack

```r
blockr_deserialize(vals$save)
$`stack-1`
[[1]]
$dataset
$value
[1] "airquality"

$choices
 [1] "airquality"       "anscombe"         "attenu"           "attitude"        
 [5] "beaver1"          "beaver2"          "BOD"              "cars"            
 [9] "ChickWeight"      "chickwts"         "CO2"              "DNase"           
[13] "esoph"            "faithful"         "Formaldehyde"     "freeny"          
[17] "Indometh"         "infert"           "InsectSprays"     "iris"            
[21] "LifeCycleSavings" "Loblolly"         "longley"          "morley"          
[25] "mtcars"           "npk"              "Orange"           "OrchardSprays"   
[29] "PlantGrowth"      "pressure"         "Puromycin"        "quakes"          
[33] "randu"            "rock"             "sleep"            "stackloss"       
[37] "swiss"            "Theoph"           "ToothGrowth"      "trees"           
[41] "USArrests"        "USJudgeRatings"   "warpbreaks"       "women"           

$multiple
[1] FALSE

attr(,"type")
[1] "literal"
attr(,"class")
[1] "select_field" "field"       
attr(,"title")
[1] ""
attr(,"descr")
[1] ""
attr(,"status")
[1] "active"
attr(,"exclude")
[1] FALSE

attr(,"name")
[1] "aoptchivyboqhey"
attr(,"expr")
get(.(dataset), envir = as.environment("package:datasets"))
attr(,"position")
[1] 1
attr(,"layout")
function(fields, ...) {
        unname(fields)
      }
<environment: namespace:blockr>
attr(,"class")
[1] "dataset_block" "data_block"    "block"        

attr(,"title")
[1] "Stack"
attr(,"name")
[1] "gghrrcxrzsmrgai"
attr(,"class")
[1] "stack"

$`stack-2`
[[1]]
$dataset
$value
[1] "airquality"

$choices
 [1] "airquality"       "anscombe"         "attenu"           "attitude"        
 [5] "beaver1"          "beaver2"          "BOD"              "cars"            
 [9] "ChickWeight"      "chickwts"         "CO2"              "DNase"           
[13] "esoph"            "faithful"         "Formaldehyde"     "freeny"          
[17] "Indometh"         "infert"           "InsectSprays"     "iris"            
[21] "LifeCycleSavings" "Loblolly"         "longley"          "morley"          
[25] "mtcars"           "npk"              "Orange"           "OrchardSprays"   
[29] "PlantGrowth"      "pressure"         "Puromycin"        "quakes"          
[33] "randu"            "rock"             "sleep"            "stackloss"       
[37] "swiss"            "Theoph"           "ToothGrowth"      "trees"           
[41] "USArrests"        "USJudgeRatings"   "warpbreaks"       "women"           

$multiple
[1] FALSE

attr(,"type")
[1] "literal"
attr(,"class")
[1] "select_field" "field"       
attr(,"title")
[1] ""
attr(,"descr")
[1] ""
attr(,"status")
[1] "active"
attr(,"exclude")
[1] FALSE

attr(,"name")
[1] "uaszvwnbofervsk"
attr(,"expr")
get(.(dataset), envir = as.environment("package:datasets"))
attr(,"position")
[1] 1
attr(,"layout")
function(fields, ...) {
        unname(fields)
      }
<environment: namespace:blockr>
attr(,"class")
[1] "dataset_block" "data_block"    "block"        

attr(,"title")
[1] "Stack"
attr(,"name")
[1] "bkwtiivvmsdtxjl"
attr(,"class")
[1] "stack"

$title
[1] "My workspace"

$settings
[1] "[]"
```

JohnCoene:
You can get it if you observe on the reactive stack.

```r
server <- generate_server(stack)

observeEvent(server$stack, {
    set_works...
})
```

But this doesn't capture field changes.

JohnCoene:
Merging `main` in this branch (not committed) with the latest changes to the workspace I'm able to save and restore fully-functional stacks! Thanks @nbenn and @DivadNojnarg

nbenn:
@DivadNojnarg at some point you had (very basic) UI support for this. Could you maybe add this here? Some button at the workspace UI/server level to trigger serialization and download the resulting json or so? Then we can merge this into mainline.

DivadNojnarg:
@nbenn: sure I can bring it back.

DivadNojnarg (approved):
LGTM. Thanks!

PR #132, opened 2023-12-15
Handle "remove" button namespace to stack level + ui tweaks

DivadNojnarg:
- This PR ensures that the block remove button is provided by the stack so that it is easier to remove a block on the server side.
- It also starts to reorganise the stack and block UI into smaller units. (Customisation will be handled separately since I have no good idea yet).
- It removes external dependencies such as `bmsui` and `sortable`.
- Collapsible triggers have been moved from the block header next to the component they control. This avoids unecessary coupling between functions and to have to control 2 pieces of UI.

@JohnCoene : as discussed please let me know if this breaks things on your side.


NAMESPACE
```diff
@@ -158,6 +159,7 @@ export(summarize_block)
 export(switch_field)
 export(uiCode)
 export(uiOutputBlock)
+export(ui_fields)
```
nbenn, line 162: Fine for now, but at some point we might have to consolidate naming a bit. To me it was not immediately clear what `ui_fields()` does. I guess what this is, is sort of `generate_ui()` for the set of fields of a block. Random thoughts:

- We could introduce an object `fields` that is a container for a set of fields and `ui_fields()` could then be `generate_ui.fields()`, the problem being that you can then no longer customize this an a per-block class basis
- Maybe we should get rid of `ui_input()` and use the `generate_ui()` for fields as well?

This is not a change request for this PR, just something to discuss when we get the chance.

R/ui.R
```diff
@@ -11,152 +11,94 @@ generate_ui <- function(x, ...) {
...
+    class = sprintf("block-inputs %s", inputs_hidden),
+    layout(fields)
+  )
+}
 
-  block_class <- "block"
-  if (.hidden) {
-    block_class <- sprintf("%s d-none", block_class)
-  }
+block_body <- function(x, ns, inputs_hidden) {
```
nbenn, line 40: Functions `block_body()`, `block_code()`, `block_title()`, etc. (or parts thereof) could be generics. This would (in parts at least) address your comment

> Customisation will be handled separately since I have no good idea yet

in a canonical manner.

The key here is to separate things that we require for stuff to work on our end, such as certain HTML class attributes etc., from things the user might want to customize. And to do this in a way that the user does not have to provide the stuff that we require in their own implementation.

This also is not a change request. It's fine as is for now, I just wanted to make the comment to sketch how we could extend customizability here.

R/server.R
```diff
@@ -63,7 +65,7 @@ generate_server.data_block <- function(x, id, ...) {
 
       # Cleanup module inputs (UI and server side)
       # and observer
-      observeEvent(input$remove,
+      observeEvent(remove(),
         {
```
nbenn, line 69: Could we maybe have this in a separate generic, such that we can provide a default implementation for block cleanup which should work in most cases, but if the block requires additional things to be finalized, that this can be handled in a block-specific cleanup function? Maybe we have something like `shiny_cleanup()` and have implementations for fields, blocks, stacks etc.? Happy to discuss this separately.

R/server.R
```diff
@@ -63,7 +65,7 @@ generate_server.data_block <- function(x, id, ...) {
 
       # Cleanup module inputs (UI and server side)
       # and observer
-      observeEvent(input$remove,
+      observeEvent(remove(),
```
nbenn, line 68: This remove trigger is supplied and some code executed at the end. In principle we might be able to make this that we wrap the module provided by the block implementer and they do not have to worry about triggering cleanup at all, no? To discuss.

nbenn: Why does the data block behave differently in this is handled from transform and plot blocks? Have they not been updated? Or is this intentionally so? Could we align this?

R/server.R
```diff
@@ -263,15 +265,57 @@ generate_server.stack <- function(x, id = NULL, new_blocks = NULL, ...) {
...
-        vals$remove <- TRUE
+      # Remove block from stack (can't be done within the block)
+      session$userData$is_cleaned <- reactiveVal(FALSE)
+      lapply(x, \(b) {
+        id <- attr(b, "name")
+        observeEvent({
+          input[[sprintf("remove-block-%s", id)]]
+          req(session$userData$is_cleaned())
+        }, {
+          # We can't remove the data block if there are downstream consumers...
```
nbenn, line 280: Things like this we should relax I feel. What is currently blocking us from doing so? Isn't this the same thing as a select block removing a column the next block might depend on? The next block this is "invalid" and down-stream eval stops until the problem is fixed.

nbenn (commented):
I added a couple of comments. None are blocking. It might be a good idea to go over them quickly in a meeting. Generally speaking I think this very much goes into the right direction. There might be some smaller things to smooth out, but we can also do so in separate PR(s) not to blow this one up (and to get things merged a bit quicker).

NAMESPACE
DivadNojnarg: We can handle this in another PR. I feel this requires much more thinking.

R/ui.R
DivadNojnarg: This is straightforward to create the generics. I can do it for this PR and we can improve later.

R/server.R
DivadNojnarg: > Why does the data block behave differently in this is handled from transform and plot blocks? Have they not been updated? Or is this intentionally so? Could we align this?

this should be cleaned. If we relax the last point (removing first block, we don't need it anymore).

R/server.R
DivadNojnarg: PR #119 does this cleanup in a better way + more unit tests. I propose to merge this one. I'll rename #119 accordingly.

R/server.R
DivadNojnarg: I relaxed it in my latest commit. I actually removed the block cleanup which will be adressed in #119.

nbenn (approved):
Thanks for addressing some of the comments. I like the directions we're taking here!

PR #135, opened 2023-12-21
Optional server modules for fields

christophsax:
WIP, will clean up and do a proper PR with tests etc.

- Optional `generate_server` method for *fields*, which makes it possible to create more complex fields
- Example `numeric_field2`: a demo of a field. Also updates UI from the server.
- Example `keyvalue_field`, as used in current mutate_block.
- mutate_block2: Alternative to mutate_block: uses `keyvalue_field` and is much closer to standard blocks.
- summarize_block2: Alternative to summarize_block: uses `keyvalue_field` to create a summarize block that is very close to mutate.


christophsax:
Abandoning, in favor of #138

Issue #136, opened 2023-12-21
Block UI issues


JohnCoene:
I don't thing this line is correct

https://github.com/blockr-org/blockr/blob/cbb1b67c23f42ae6333e760a58e2c62f8c3e2813/R/ui.R#L235

This prepends to `.block-tools`, it does not place it within the `div`

![image](https://github.com/blockr-org/blockr/assets/7833796/78724b02-2a3f-4e26-afc5-8e9d68d2a006)

Also, shouldn't the `.block-tools` be within the parent `flex`?

DivadNojnarg:
I spotted another issue while playing with dynamic blocks. The delete button is missing because it is added only from the stack once. 

<img width="1035" alt="Screenshot 2023-12-21 at 17 03 04" src="https://github.com/blockr-org/blockr/assets/18291543/374336e5-71be-4a32-8987-739f8cca004e">

JohnCoene:
The show code at the block-level has also disappeared

JohnCoene:
Closing this I think it's been tackled, we can open separate, more focused, issues if anything else is still to do

PR #137, opened 2023-12-29
fix: collpase body placed in header #136

JohnCoene:
This addresses some of the issues in #136 

We're still missing the code output at the block level, I'm not sure why this was removed.


DivadNojnarg:
Thanks for checking this. The collapse triggers should be located as shown below in red. The idea was to have the collapse trigger and the collapsible element in the same function for easier customisation (before we had all the triggers in the block head and the collapsible targets in different components). Do you mind changing the icon so they look like collapse triggers? I kept the original icon as they were in the first iteration but now this is misleading. Feel free to improve the design if you wish so we have a better separation between the code output and the block main output. 

292256518-374336e5-71be-4a32-8987-739f8cca004e

I'll review when I am back next Wednesday.

JohnCoene:
OK, I did not know we were changing all the UI.

It's currently very confusing.

DivadNojnarg:
Side note: code to add the collapse code button within the header from the stack method. This however re-introduces the decoupling issue we had before (trigger and target in 2 different pieces of code, introducing some fragility if people had to customise the block_body or block_code methods). 

```r
# within generate_ui.stack

collapse_code_id <- htmltools::tagQuery(tmp)$
            find(".block-code")$
            selectedTags()[[1]]$
            attribs$id

          # Remove button now belongs to the stack namespace!
          htmltools::tagQuery(tmp)$
            find(".block-tools")$
            prepend(
              list(
                tags$a(
                  class = "text-decoration-none block-output-toggle",
                  href = sprintf("#%s", ns("outputCollapse")),
                  iconOutput()
                ),
                tags$a(
                  class = "text-decoration-none block-code-toggle",
                  `data-bs-toggle` = "collapse",
                  href = sprintf("#%s", collapse_code_id),
                  `aria-expanded` = "false",
                  `aria-controls` = collapse_code_id,
                  iconCode()
                ),
                block_remove(b, ns(sprintf("remove-block-%s", block_id)))
              )
          )$
          allTags()
```

JohnCoene:
@DivadNojnarg I cannot see this snippet in main

DivadNojnarg:
That's something I checked this morning but did not push. I keep it here to remember for later, if we use it.

PR #138, opened 2024-01-03
optional generate_server for fields

christophsax:
Basic infrastructure to support fields with own server module.

### Basic Idea

- If a field has a generate_server() method, initialize it and use
  its return value.
- If not, use existing code
- combine both to r_values()
- to update ui, use ui_update(), or update module init reactive value

So far, this is implemented in `generate_server.transform_block()` only. If ok, I can do the same for `plot_block()` and `data_block()`, which is a copy and paste exercise.

`numeric_field2` is just for demonstration. If you like, I can exclude it from this PR.

### After this PR

If this PR is ok, I can add new fields and blocks that make use of this:

- A new mutate block, that uses a `keyvalue_field` and does not need its own server logic.
- A new summarize block, that uses the same `keyvalue_field` as mutate and creates a more flexible summarize experience
- A new admiral demo block, that uses an arbitrary expression field.


christophsax:
All checks green now

nbenn:
@christophsax I'm not sure how to best proceed here. The essence of this PR is a change to `generate_server.transform_block()`. All else is purely for demo/test purposes, no? As long as the PR does not contain useful applications of the new functionality, why not structure the PR as such? What I mean is the following:

- Add the necessary changes to all block-level `generate_server()` implementations (any maybe figure out what to abstract into re-usable functions for further block-level `generate_server()` implementations).
- If you need an example for a block/field for a test, create this field as part of the test (instead of adding exported functionality as `numeric_field2()`/`head_block2()` which we probably do not want to do this way anyways and will have to be removed again)
- (Or if you feel that `numeric_field()`/`head_block()` should be changed, do so there)

christophsax:
Yes, all the rest is just demo. I am fine with changing this to a pure `generate_server.transform_block()` update and removing the demo.

A more interesting example (updated keyvalue fields) comes in another PR. 

Will update later today.

christophsax:
@nbenn PR #146 does essentially the same, but uses the `keyvalue_field` in an updated `mutate_block` and implements a test for that. That way, I no longer need the demo field and block. Does it make sense to merge this instead?

nbenn:
@christophsax do we want to close this and focus on #146? Btw I did not mean for you to completely remove the code for `numeric_field2()`/`head_block2()` from the repo entirely, just form the main `R` folder. Those are useful *examples* on how server-based fields could be used and as such would be great as test cases, no?

christophsax:
Yes, let's close this. I can re-add the examples later on to a meaningful location.

PR #139, opened 2024-01-03
Dynamic removal of blocks

DivadNojnarg:
Will fix one of the issue in #136, where the remove button is missing for dynamically added blocks.


DivadNojnarg:
@nbenn : unit tests will follow in another PR. I already have some of them in #119 and don't want to copy and paste here.

DivadNojnarg:
Note @nbenn: I can also make `handle_remove_block` an internal function instead of S3, if we assume people won't need to customise it, which is likely.

nbenn:
> Note @nbenn: I can also make `handle_remove_block` an internal function instead of S3, if we assume people won't need to customise it, which is likely.

@DivadNojnarg You could also not export the generic function itself I believe. The corresponding (class-specific) methods have to be exported, but as long as the generic is not, it should behave as "interal" function, despite using s3 dispatch.

nbenn (commented):
No objections from my side.

DivadNojnarg:
Great. I am done with this PR. This can go to main.

nbenn (approved):
Lgtm.

PR #140, opened 2024-01-03
Validation: not hardcoding colors

JohnCoene:
This PR removes hardcoded styles for validation, namely border and text colors.

Note:

I see we map over `ids`.

https://github.com/blockr-org/blockr/blob/cbb1b67c23f42ae6333e760a58e2c62f8c3e2813/srcjs/shiny.js#L62

But this is sent from here:

https://github.com/blockr-org/blockr/blob/cbb1b67c23f42ae6333e760a58e2c62f8c3e2813/R/utils.R#L325

I don't understand how `ns(el)` can return a `list`


DivadNojnarg:
@JohnCoene: It's possible that I forgot to cleanup the JS code after rearranging the R part. I'd suggest to remove the `msg.id.forEach` and the if condition, because it does not make sense anymore looking at the R part.

JohnCoene:
I had removed the `if` statement but this last change removes it all and assume `id` is always and `String`

DivadNojnarg:
I tested the branch. I see the block border color when invalid but I don't see the individual input colors. My code:

```r
stack <- new_stack(
    data_block,
    filter_block
)
```
The text input should have a red border like the block when empty. On a side note, If you manage to make it work for any kind of input that would be great! (so far my old code does not work for selectInput or multiselect).

JohnCoene:
Sorry something wrong with my last revert.

Fixed now.

It works for select inputs too

srcjs/shiny.js
```diff
@@ -44,7 +44,7 @@ const changeInputBorder = (args) => {
   // Some inputs are dynamically generated like in filter block.
   // Adding a delay ensure they're in the DOM.
   setTimeout(() => {
-    if (state) {
+    if (args.state) {
```
DivadNojnarg, line 47: Shouldn't this be `!args.state` instead? At the moment, I get a valid message when the input is empty and an error when the input if filled.

JohnCoene: Indeed, fixed.

DivadNojnarg (approved):
Good to merge. Thanks.

PR #141, opened 2024-01-03
Show inputs on error

JohnCoene:
This PR ensures we display the block inputs on error (as reported by validation).

e.g.:

```r
stack <- new_stack(
  data_block,
  filter_block
)
```

This results on a validation error.
On load the error message is displayed but the inputs are not visible.
This is because on load we hide inputs and only display the last output (table, plot, etc).
But if the validation fails there are no outputs to show (expr not evaluated) and the inputs are hidden which is confusing.

This PR fixes this: if the an error is reported we ensure we display the inputs.


DivadNojnarg (approved):
Yes this is better for the end user if the pipeline has many blocks. Thanks.

PR #145, opened 2024-01-03
Make title editable fixes

JohnCoene:
This makes the title editable.

It uses `setInputValue` instead of a custom registered input since it should not be needed anyways.


srcjs/stack-title.js
```diff
@@ -0,0 +1,30 @@
+export const title = (stack, ns) => {
+  const $title = $(stack).find(".stack-title");
+
+  $title.on("click", (e) => {
+    console.log(e);
```
DivadNojnarg, line 5: remove when approved.

R/server.R
```diff
@@ -369,6 +362,17 @@ generate_server.stack <- function(x, id = NULL, new_blocks = NULL, ...) {
         )
       })
 
+      observeEvent(input$newTitle, {
+        attr(vals$stack, "name") <- input$newTitle
```
DivadNojnarg, line 366: attr(vals$stack, "name") is what we use to initialise the stack namespace if no ID is set when calling `generate_server.stack`. Blocks inputs inherit from this. Do you see any bad consequence of doing this on the serialisation and restoration of a dashboard?

DivadNojnarg (commented):
See my question below. If there is no impact, we are free to go.

R/server.R
JohnCoene: It's also used for namespace of a lot of things downstream.
It may work but I'd rather not take a chance.

How can we make the title editable otherwise.

I don't think we should use the id/namespace for display.

srcjs/stack-title.js
JohnCoene: Sorry, removed

JohnCoene:
See, last commit, would that work? [5f6364e](https://github.com/blockr-org/blockr/pull/145/commits/5f6364e4c93e40e6b9d2a0b43840b71f39b9e7c0)

Essentially we keep name but it's just for internal use, we rename the exposed argument `title`.

@nbenn let me know if this is ok with you

I looked at the `serialise` branch at it looks like it should not impact the `stack` serialiser.

https://github.com/blockr-org/blockr/blob/1b6fbd2eb32a0e18e792676a7756c1308fbe4236/R/serialise.R#L28

R/server.R
DivadNojnarg: > Essentially we keep name but it's just for internal use, we rename the exposed argument title (rename in this commit [ccd469b](https://github.com/blockr-org/blockr/pull/145/commits/ccd469bf1f84655ed1b1da94b6490e74c1124228))

I'd actually also rename `name` to `id` so we don't make the confusion, especially that would align with the optional `id` parameter exposed by the `generate_server.xxx` methods.

R/server.R
```diff
@@ -369,6 +362,17 @@ generate_server.stack <- function(x, id = NULL, new_blocks = NULL, ...) {
         )
       })
 
+      observeEvent(input$newTitle, {
+        attr(vals$stack, "title") <- input$newTitle
```
nbenn, line 366: @JohnCoene How about adding function `set_stack_title()` that maybe ensures that the input is a string and lives somewhere in `R/stack.R`? Just to try to not rely on implementation details and try building an API for manipulating our objects instead.

R/ui.R
```diff
@@ -254,9 +254,9 @@ stack_header.stack <- function(x, title, ns, ...) {
     div(
       class = "d-flex",
       div(
-        class = "flex-grow-1",
+        class = "flex-grow-1 d-inline-flex",
         span(icon("cubes"), class = "text-muted"),
-        title
+        span(attr(x, "title"), class = "stack-title cursor-pointer")
```
nbenn, line 259: See previous comment and maybe add a `get_stack_title()` function.

test.R
```diff
@@ -5,7 +5,8 @@ library(shiny)
 stack <- new_stack(
   data_block,
   select_block,
-  filter_block
+  filter_block,
+  name = "test"
```
nbenn, line 9: Should this not be `title`?

nbenn (commented):
No substantial comments, I see no problems, as the state of the stack object is properly updated.

test.R
JohnCoene: That's in the following commit

R/server.R
JohnCoene: I agree but I suggest we do that in a separate PR because this has quite some impact on the codebase, e.g.: breaks use of `names()`

DivadNojnarg (approved):
Thanks, LGTM. We'll change the name to id in another PR.

DivadNojnarg:
@JohnCoene: There are few conflicts now but this is good to merge after fixing them.

JohnCoene:
Fixed conflicts

PR #146, opened 2024-01-04
Key value field / streamlined mutate block

christophsax:
Builds on PR #138 

- adds a new `keyvalue_field` (subsequent PRs will improve this)
- uses the `keyvalue_field` in a mutate block that looks now like normal blocks
- A test for the mutate block covers the essence of PR #138, so the demo fields `head_block2` and `numeric_field2` are no longer needed and have been removed


christophsax:
@nbenn Probably preferable to #138

nbenn:
@christophsax I tried using a keyvalue field as

```r
new_df_block <- function(...) {

  handle_error <- function(e) {
    warning(e)
    expression()
  }

  parse_one <- function(x) {
    tryCatch(
      parse(text = x),
      error = handle_error
    )
  }

  parse_kv <- function(columns) {
    bquote(
      data.frame(..(exprs)),
      list(exprs = lapply(columns, parse_one)),
      splice = TRUE
    )
  }

  fields <- list(
    columns = new_keyvalue_field(),
    expression = new_hidden_field(parse_kv)
  )

  new_block(
    fields = fields,
    expr = quote(.(expression)),
    ...,
    class = c("df_block", "data_block")
  )
}

df_block <- function(...) {
  initialize_block(new_df_block(...))
}

serve_stack(new_stack(df_block))
```

but the result is buggy. Am I doing something wrong? Can you reproduce?

https://github.com/blockr-org/blockr/assets/3158446/243b5dff-5ad1-4672-be38-5de3d5cf447b

christophsax:
Ah.. **currently only supports `transform_block`**. Bringing this to `data_block` and `plot_block` is a copy paste exercise. Can do this afterward.

christophsax:
@nbenn, the second idea from above is done in the last commit. Now your data example should run, too.

nbenn:
@christophsax does it work for you now? For me it doesn't. I can add/rm fields, but neither is the expression updated nor do I see any tabular output. I believe my `parse_kv()` function is not executed, but did not do any proper debugging yet. (Might also be my fault). Another observation: in order to see stuff, quite some clicking is necessary in the beginning. Is the UI being initialized properly?

https://github.com/blockr-org/blockr/assets/3158446/e08ec4c8-e795-4727-9dba-23243d93bbaa

nbenn:
@christophsax starting with a simple example also causes an issue for me:

```r
serve_stack(new_stack(data_block))
```

does not seem to do anything with the current server architecture.


https://github.com/blockr-org/blockr/assets/3158446/1a7096a0-ef9d-4fc5-84d4-ccfed15fabfc

nbenn:
@christophsax another small experiment is also causing me issues: I tried doing a mutate with two "expression" fields, one for the `...` and one for the tidyselect enabled `.by` argument. The behavior of the "Submit" buttons is a bit funky.

```r

  handle_error <- function(e) {
    warning(e)
    expression()
  }

  parse_one <- function(x) {
    tryCatch(
      parse(text = x),
      error = handle_error
    )
  }

    res <- lapply(x, parse_one)

    if (length(res)) {
      res <- do.call(c, res)
    }

    res
  }

    parse_multiple(new_cols)
  }

    names(res) <- NULL
    res
  }

  fields <- list(
    new_cols = new_keyvalue_field(value = list()),
    group_by = new_keyvalue_field(value = list()),
  )

  expr <- quote(
  )

  new_block(
    fields = fields,
    expr = expr,
    ...,
  )
}

}

  do.call(
    bquote,
    list(
      attr(x, "expr"),
      where = lapply(x, value),
      splice = TRUE
    )
  )
}

    fun <- x[[cmp]]
  }

  x
}

}

```

Some observations

- initially the 2nd "Submit" does not do anything
- choosing the first one updates as expected
- on subsequent changes, only the button where the change was made, updates the view
- if several field groups are changed, submit buttons have to be chosen individually 

https://github.com/blockr-org/blockr/assets/3158446/e949467e-1c8b-44e4-b28b-ae3e399d339e

I don't think it's a good idea to have "Submit" functionality at the individual field level. This should be an action that (optionally) is provided by the block and acts on all fields together.

(The function `eval_field()` has to be used for init due to #156. This should be looked into separately.)

nbenn:
@christophsax I wanted to try introducing a dependency (on another field or the data) and a test-case here could be auto-complete support for columns. Currently this does not seem to be supported (right?). The kv field constructor would need something like an argument for additional auto-complete values (maybe a character vector?) that could be passed to the ace editor.

christophsax:
@nbenn latest commit fixes the first two issues and adds a test for the first issue.

I don't think the exact look and feel of the keyvalue field options is too important here (I have a follow-up PR, #155, that introduce an option to turn off the submit button). We can easily iterate on that later on and add, e.g., options for autocomplete.

nbenn (commented):
I have said my piece.

christophsax:
@nbenn could you approve or merge? The `codecov/project` check comes from the fact that `server.R` has less lines now, due to the out factoring.

nbenn:
@christophsax I cannot merge either. Feel free to disable the main branch protection under 

https://github.com/blockr-org/blockr/settings/branches

and push directly to main. Looking at your diff, I believe that you're trying to re-introduce quite a bit of (removed?) stack server code which therefore probably is untested. Not sure, what exactly your plan is there and whether this is part of your "out factoring".

DivadNojnarg (approved):

DivadNojnarg:
@christophsax: this should be mergeable now

christophsax:
Thanks! @nbenn, yes, you are right, I missed some stack updates here. Will resolve.

DivadNojnarg:
Regarding the coverage, we can extract tests from https://github.com/blockr-org/blockr/pull/119 which covers most of the server testing (stack, ...). I opened an issue https://github.com/blockr-org/blockr/issues/170 to do this.

christophsax:
Seems to be good now. @nbenn is not a big fan of a server logic for fields, but it makes field creation much easier for me. Let's see where we go with this in the long run. Thanks.

PR #147, opened 2024-01-04
Rearrange buttons on block UI

JohnCoene:
This rearranges buttons on block UI so it looks like it did originally: placing the buttons on the top right.

Note that now that the code block is within the body of the block the button cannot be shown when the block is collapsed.


DivadNojnarg (commented):
Thanks @JohnCoene. I am ok with the show code not always visible.

Few requests:

- Can we change the collapse icon to look like more https://fontawesome.com/icons/chevron-up?f=classic&s=solid and https://fontawesome.com/icons/chevron-down?f=classic&s=solid dynamically. 
- Can we add a tooltip for each button ("Show code", "Collapse block", ...)?

Same at the stack level:

- Replace the edit button by a dynamic collapse button. 
- Show a tooltip on the code icon.

JohnCoene:
Updated icons.
Just added tooltip to copy code button on stack, block already has `data-bs-toggle` so it's a bit more work and I have other more urgent things to tackle before.

DivadNojnarg (changes requested):
Thanks for adding the new icons. There seem to be an inconsistency from a block to another (see below): 2 blocks in the same state but the icon isn't the same.
Screenshot 2024-01-04 at 16 54 37

JohnCoene:
I cannot reproduce this, can you tell how you got to that?

JohnCoene:
Reproduced, fixed in last commit

DivadNojnarg (approved):
Thanks. LGTM.

PR #155, opened 2024-01-05
Provide options for keyvalue_field / use in filter2_block

christophsax:
It's a small PR once #146 is merged.

- Provide options for `keyvalue_field`
- Use in `filter2_block`
- Suggest to include `filter2_block` and discuss later if we want to substitute the current `filter_block` by it.


christophsax:
Abandoning, in favor of #189

PR #161, opened 2024-01-09
Prevent cyclic block updates

### LINKED ISSUE #159: Block level update trigger
The `observeEvent()` for our block-level update mechanism

https://github.com/blockr-org/blockr/blob/1c5c1c87e9d67323ddf383d612102c84486b90b7/R/server.R#L120-L127

might have been the wrong choice. Instead, we could do something like

```r
reactive({
   Sys.sleep(2)  # Pretend this is an expensive computation
   input$x * input$y
 }) %>%
 bindCache(input$x, input$y) %>%
 bindEvent(input$go)
```

to ensure that only *value changes* trigger updates and the only relevant *event trigger* is an (optional) block level "submit" button.

I will try to create an example where the current design causes an infinite loop of updates.

nbenn:
This is intended to fix #159.

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -41,7 +41,7 @@ generate_server.data_block <- function(x, id, ...) {
       blk <- reactiveVal(x)
 
       o <- observeEvent(
-        eval(obs_expr(blk())),
+        eval(obs_expr(isolate(blk()))),
         eval(set_expr(blk())),
         ignoreInit = TRUE
       )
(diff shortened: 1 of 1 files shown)


nbenn:
@DivadNojnarg how does this look to you? I believe this removes some of the unneeded executions of our update routines. At least it solves the cyclic updates for me.

At the time when we added

https://github.com/blockr-org/blockr/blob/1c5c1c87e9d67323ddf383d612102c84486b90b7/R/server.R#L120-L127

I was under the impression that this would not incur a dependency on `blk` but now it seems to me that it does, which would be unwanted.

As part of this PR I would also address #162.

nbenn:
Closed in favor of #146.

PR #169, opened 2024-01-09
Icon toggle

JohnCoene:
This fixes the icon toggle and the initialisation of new blocks

### DIFF
--- a/test.R
+++ b/test.R
@@ -10,7 +10,6 @@ stack <- new_stack(
 )
 
 ui <- fluidPage(
-  "test",
   theme = bslib::bs_theme(5L),
   generate_ui(stack)
 )


DivadNojnarg:
I just noticed something weird with the delete button.

https://github.com/blockr-org/blockr/assets/18291543/766505a5-a00c-4aa4-9392-53a82fd03d23

JohnCoene:
This is also caused by the recent changes in the block ui: the remove button is rendered within `.block-tools` which is within the `.block-title` which is hidden when collapsed.

JohnCoene:
Note: I also pushed a quick fix to the lock. I was importing/using the wrong function.

JohnCoene:
Should be good to review/merge. I can look at any other issue in separate PRs

JohnCoene:
Ready for review

DivadNojnarg (approved):
Remove button visibility will be fixed in another PR since not related to any change in this PR.

Issue #165, opened 2024-01-09
Filter block: validation error on add


JohnCoene:
When we add a filter_block to an existing tag we are prompted with this:

![image](https://github.com/blockr-org/blockr/assets/7833796/d955aba1-83f6-4835-9df6-cbc2d95908ae)

Also, I'm not sure we should really interrupt the user with a modal, there is already ample feedback when there is an error in a block.

DivadNojnarg:
To remove from `secure` in utils.R

DivadNojnarg:
This error is not relevant anymore in main because `secure` is not being used when attempting to update a block (therefore no error is thrown).

PR #171, opened 2024-01-10
Empty Stack

JohnCoene:
This initialises work on allowing for an empty stack.

@DivadNojnarg it admittedly requires some refactoring but let me know what you think.

### DIFF
--- a/R/stack.R
+++ b/R/stack.R
@@ -11,6 +11,10 @@ new_stack <- function(..., title = "Stack") {
   ctors <- c(...)
   names <- names(ctors)
 
+  if (!length(ctors)) {
+    return(structure(list(), title = title, name = rand_names(), class = "stack"))
+  }
+
   blocks <- vector("list", length(ctors))
 
   blocks[[1L]] <- do.call(ctors[[1L]], list(position = 1))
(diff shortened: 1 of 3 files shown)


R/stack.R
```diff
@@ -61,32 +65,40 @@ generate_code.stack <- function(x) {
 #' @return Invisibly returns the stack.
 #' @export
 add_block <- function(stack, block, position = NULL) {
-  stopifnot(length(stack) > 0)
   if (is.null(position)) stopifnot(position >= 1)
```
DivadNojnarg, line 68: I believe we should remove this line.

JohnCoene:
Sorry for being pushy but could we look at this PR, I really need to be able to initialise empty stacks in order to proceed with other tasks.

DivadNojnarg:
I am trying to test it but something is broken. Now the passed block to `add_block` is a function

```r
$ block
function(...) {
  initialize_block(new_data_block(...))
}
```

 and not a structure which crashes on the first check:

```r
Error in add_block: The first block must be a data block.
```

R/stack.R
```diff
@@ -61,32 +65,40 @@ generate_code.stack <- function(x) {
...
   if (length(stack) == 0) {
-    block_name <- deparse(substitute(block))
```
DivadNojnarg, line 67: This is what causes the issue with the block class check because block is a function.

JohnCoene:
That's what I get from @nbenn registry `available_blocks()`, e.g.: `new_block <- reactive({available_blocks()[1}[[1]])`

How should I go about it?
It's confusing because the registry doesn't return a list of blocks, these are `block_descr`

R/stack.R
JohnCoene: How do you use/test this, could you share an example?

DivadNojnarg: @JohnCoene I just pushed a fix for the example. I noticed another issue, but this is actually not related to this PR. It seems that blocks in the registry don't call the `initialize_block` (`new_data_block` and `data_block` are different, the latter calling `initialize_block(new_data_block(...))`) method that's why you get an error why you insert a filter block. This is explains #172 and #165.

DivadNojnarg (approved):
LGTM.

R/stack.R
JohnCoene: Ah good find!

JohnCoene:
Thanks David!

Issue #173, opened 2024-01-11
Remove/add block broken with latest PR


DivadNojnarg:
It seems that this function has been removed in the latest PR #146, the conflict resolution isn't correct: 

See https://github.com/blockr-org/blockr/pull/146/commits/b4e2ac98ed7b0c25ea86b4702d3cbb04c4e5c228#diff-9e95b340633529fb09e594e8e485b849dac8a3aa0d81017be4c7556d34bfb05bL252-L283.

We also can't add new block because `handle_remove_block` is called within the stack but not defined.

```r
Error in handle_remove_block: could not find function "handle_remove_block" 
```

The app is currently unusable (cc @JohnCoene). Can you revert this @christophsax? I did not check everything but there might be other problems.

DivadNojnarg:
The last valid state of the app is: https://github.com/blockr-org/blockr/tree/a985dc9c56595493349045a759a52d366da317d5

christophsax:
We can revert my PR. I think I haven't merged it properly.

christophsax:
I did a PR that brings the function back (#175). Does that fix the problem?

PR #175, opened 2024-01-11
Bug/accidental remove handle remove block

christophsax:
@DivadNojnarg does that fix https://github.com/blockr-org/blockr/issues/173?


DivadNojnarg (commented):
I can delete/add blocks. What was the reason to remove the validation?

JohnCoene:
Validation and the improvements we made to it seem indeed to be gone.
Also seems to have broken some rendering of blocks where we render collapsed to only show last output.

christophsax:
No reason to remove validation. That was a shortcut I did during development. Will re add it.

christophsax:
@JohnCoene will push a fix for the validation in half an hour.

Can you give me an example for the rendering issues?

Sorry for the trouble, but my PR enables us to follow up with a range of useful blocks so I would be happy if we don't have to revert.

DivadNojnarg:
@christophsax: what about keeping your changes but restoring the old `generate_server` methods as we had before and making the code depend on them again. This will give space to follow up and make sure everything runs smoothly when ready.

JohnCoene:
Is #178 related to this?

christophsax:
@DivadNojnarg, replacing the old `generate_server` methods (for blocks) was the central piece of the PR. I haven't intended to touch the method for the stack. If you want to revert the methods for the blocks, we probably should revert everything.

But I think this PR resolves most of the problems, doesn't it?

nbenn (approved):
Lgtm.

PR #177, opened 2024-01-11
Fix: lock

### LINKED ISSUE #176: Lock - dependency issue
I have to do some re-work on the lock functionality.

The issue is that `lock`, `unlock`, `toggle_lock` can run before any `stack` is rendered so before the JavaScript dependency is imported, therefore the "lock" message is sent through the websocket 

https://github.com/blockr-org/blockr/blob/008965264e9843ff43defe4692434d81decc46e4/R/lock.R#L79

but the message handler is not present since the dependency is not attached

https://github.com/blockr-org/blockr/blob/008965264e9843ff43defe4692434d81decc46e4/srcjs/lock.js#L2

JohnCoene:
This PR fix #176 

When we "bind" a stack we send the lock state so we can appropriately render the stack.

See example below where we run `lock` before rendering a stack. 

```r
library(blockr)
library(blockr.data)
library(shiny)

stack <- new_stack(
  data_block,
  select_block,
  title = "test"
)

ui <- fluidPage(
  theme = bslib::bs_theme(5L),
  useBlockr(),
  uiOutput("stack")
)

  observe(lock())
  output$stack <- renderUI({
    generate_ui(stack)
  })
  x <- generate_server(stack)
}

shinyApp(ui, server)
```


test.R
```diff
@@ -5,16 +5,20 @@ library(shiny)
...
   data_block,
   select_block,
-  filter_block,
   title = "test"
 )
 
 ui <- fluidPage(
   theme = bslib::bs_theme(5L),
-  generate_ui(stack)
+  useBlockr(),
```
DivadNojnarg, line 13: This a detail, why not exporting `blockrDeps`? Or renaming to `useBlockr`, which is maybe more meaningful (useWaiter, ...).

DivadNojnarg (commented):
If I run your example, working well, and then, I restart R and run another simple example (not locked):

```r
pkgload::load_all()

stack <- new_stack(
    data_block,
    filter_block,
    select_block
)

serve_stack(stack)
```

the new stack opens locked.

test.R
JohnCoene: I didn't see this as a problem but anyways renamed so we can ship this asap

JohnCoene:
Sorry, pushed a fix, some `toggle()` ran on load

DivadNojnarg (approved):
Thanks for fixing this.

Issue #149, opened 2024-01-04
Data Block: read data from disk


JohnCoene:
We're going to need a block that allows reading data from disk.

This can be implemented a number of ways. The user should be able to select a dataset from a directory.

Note that we should be able to handle multiple data formats, e.g.: `csv`, `xpt`, `sas7bdat`

DivadNojnarg:
@JohnCoene: do you expect the user to also browse the directory? For instance having access to the file system while using the tool. `{shinyFiles}` does this, however, we'd have to create extra fields and so ...

JohnCoene:
Yes, being able to browse files would be good I guess.

I have a feeling that shiny module is not ideal for fields. maybe we can build something on top of [this](https://github.com/cubiclesoft/js-fileexplorer) (or any other library) instead?

What do you think?

nbenn:
@DivadNojnarg I'm for adding this as a regular new field, something like `upload_field` or so. Does not shiny provide `fileInput()` (which would be called in `ui_input.upload_field()`) out of the box? Or is that no good for some reason?

If you want to add shinyFiles as regular field, what I do in

https://github.com/blockr-org/blockr/blob/979a176435a493712d99d6741961108954b7297f/R/ui.R#L557-L585

might be helpful. You can of course also use the field-server stuff from @christophsax , but I don't believe this is necessary only for registering an observer.

If you want me to expand on what I do in the above snippet, let me know. If you cannot make the above work in the non-field-server setting, this would interest me as well.

DivadNojnarg:
Can we gather #149 and #148 in that case?

nbenn:
@DivadNojnarg sorry, I started mixing things up there. I guess `upload_field` and `file_field` can be two separate things. I did not pay close attention and thought files would be client-provided and that shinyFiles was just a fancy (client-side) file browser.

Is server-side file browsing really a requirement @JohnCoene? I know nothing about the BMS infra, but I'd imagine that such files are not actually on a server-local filesystem.

JohnCoene:
Yes, they are on the a drive mounted on the Connect Server so we need a new field/block for users to select the appropriate file from disk

nbenn:
Ok, then I guess we need both an `upload_field` and a `file_field` (or whatever better names we can come up with).

[...]

Issue #148, opened 2024-01-04
Data block: upload


JohnCoene:
We're going to need an "upload block."

Some companies do no allow data being stored on the environment where their application run (e.g.: non-GXP compliant environments)

JohnCoene:
Will require a file upload field

DivadNojnarg:
Upload would look like:

```r
new_upload_block <- function(...) {

  read_data <- function(dat) {
    if (is.null(dat)) return(quote(TRUE))
    data_func <- read.csv # TO DO switch
    bquote(
      .(read_func)(.(path)),
      list(read_func = data_func, path = dat$datapath)
    )
  }

  new_block(
    fields = list(
      dat = new_upload_field(),
      expression = new_hidden_field(read_data)
    ),
    expr = quote(.(expression)),
    ...,
    class = c("upload_block", "data_block")
  )
}
```

Note for me: the initialisation of subsequent transform block seems "broken" and I need to re-upload a dataset to get it working.

<img width="1340" alt="Screenshot 2024-01-12 at 16 15 28" src="https://github.com/blockr-org/blockr/assets/18291543/44278b74-3571-42d3-8d6b-bae8263a11e2">

nbenn:
@DivadNojnarg If you start up a PR, I'm happy to comment there. A couple things that come to mind:

- 
    > the initialisation of subsequent transform block seems "broken"
    
    Might that be b/c you're initially returning `quote(TRUE)` and this does not work for the downstream transform block? Maybe an empty data.frame instead?
    
- the `new_hidden_field` is used quite pervasively and I believe in some cases it is not necessary. Here you could do the `read_func` as `select_field`, but that still leaves you with the path only being available as `dat$datapath`, so you still need a tiny bit of custom logic. One thing that we have rarely used so far, but could lead to a bit more straightforward code is providing a block-specific `generate_code()` method instead of relying on the default block level version. By providing a `generate_code.upload_block()` method you can assemble your expression a bit easier than `bquote()`ing stuff together in a `hidden_field`. Just a thought. Try it out and let me know what you think[^1].

[^1]: I have been playing with the thought of maybe also allowing passing of a function as `expr` in `new_block()` b/c (1) sometime a tiny bit of processing is required (e.g. extracting `dat$datapath`) but (2) providing that logic as `hidden_field` is not a very elegant solution to say the least.

nbenn:
[...]

PR #184, opened 2024-01-12
Improve data loading block

DivadNojnarg:
See #148, #149.


DivadNojnarg:
TO DO:
- find a way to allow user to pass in different browsable folders with the `filesbrowser_field`. Right now, I hardcoded it to `volumes <- c(home = path.expand("~"))` for testing.

DivadNojnarg:
@nbenn : as I couldn't make any of the upload/filebrowser field properly render within `variable_field`, we temporarily have 2 handle 2 blocks (1 block containin the 2 field would be too messy). That should be good enough for basic usage though.

DivadNojnarg:
I noticed `ignoreInit = TRUE` is causing any subsequent block to `upload_block` not to update properly (have to re-upload data). That refers to one of my comment in #148.

christophsax:
@DivadNojnarg re `ignoreInit = TRUE`, are you referring to this? #196 seems good to me, but not sure if we overlooked something.

PR #189, opened 2024-01-14
Options and Tweaks for key_value field

christophsax:
This should have no impact on non-server fields. The only exception is a `ignoreInit = FALSE` [change](https://github.com/blockr-org/blockr/pull/189/commits/e30b6f06f908fcfd9e7a617c2d8f6b833186a4e2), which is now needed for correct initialization.

### Improved `key_value field`

- Now uses `updateUI`, instead of `addUI` etc.
- Respects initialization value. E.g., if you define `new_filter2_block <- function(data, value = "Time == 2", ...) {`, this value would appear in a new block. That should make it work with serialization, too.
- Uses colnames from data
- Should be generally much more stable
- Adds 'filter2' block, as an expression-based alternative for our current filter block.


nbenn (approved):
As with #146, I don't think this is mature enough for merging. With that said, I'm not standing in your way.

nbenn:
@christophsax a variant of the test case in https://github.com/blockr-org/blockr/pull/146#issuecomment-1879684292 (using new `keyvalue_field` arguments) is the following

```r

  handle_error <- function(e) {
    warning(e)
    expression()
  }

  parse_one <- function(x) {
    tryCatch(
      parse(text = x),
      error = handle_error
    )
  }

    res <- lapply(x, parse_one)

    if (length(res)) {
      res <- do.call(c, res)
    }

    res
  }

    parse_multiple(new_cols)
  }

    names(res) <- NULL
    res
  }

  suggest_by <- function(data) colnames(data)[1L]

  fields <- list(
    new_cols = new_keyvalue_field(value = list()),
    group_by = new_keyvalue_field(suggest_by, multiple = FALSE, key = "none"),
  )

  expr <- quote(
  )

  new_block(
    fields = fields,
    expr = expr,
    ...,
  )
}

}

  do.call(
    bquote,
    list(
      attr(x, "expr"),
      where = lapply(x, value),
      splice = TRUE
    )
  )
}

    fun <- x[[cmp]]
  }

  x
}

}

```

christophsax:
Thanks, will look at it.

christophsax:
Abandoned in favor of #203

PR #190, opened 2024-01-15
Stop using internal function in user facing demo app

### LINKED ISSUE #186: Using internal function in demo
`Error in add_block_ui(identity) : could not find function "add_block_ui"`

TO DO:

- export `add_block_ui`.

- In `server.R`, `list_blocks()[[as.numeric(input[[sprintf("%s-selected_block", id)]])]]` is now invalid and should use `available_blocks()`:

```r
inject_block <- function(input, vals, id) {
  observeEvent(input[[sprintf("%s-add", id)]], {
    # Reset to avoid re-adding existing blocks to stacks
    vals$new_block <- NULL
    block <- list_blocks()[[as.numeric(input[[sprintf("%s-selected_block", id)]])]]
    # go at the end)
    vals$new_block[[id]] <- list(block = block)
  })
}
```

DivadNojnarg:
DONE: Export add_block_ui + refine doc + use correct registry function in server

To fix #186 which prevents one the demo app from running (add block demo).

Good for review @nbenn


nbenn (approved):
One thing that might be considered (for this PR or a future one) is `construct_block()` for block construction. A `block_descr` is basically a function with some attributes, but in the spirit of not relying on such implementation details, we have (and if needed, should adapt) an API for interacting with these objects.

DivadNojnarg:
I'll open a new issue for this improvement. The intent here was more a hotfix so the demo app works again.

PR #194, opened 2024-01-16
Rendering blocks collapsed

### LINKED ISSUE #192: Block render
A previous PR seems to have broken some of the rendering of the blocks, namely rendering as collapsed.

JohnCoene:
This fixes #192 

We need to render blocks collapsed since we only need to present the outputs of the blocks on load.
However, we need to render the blocks expanded when they are dynamically added from the UI we want the blocks to be expanded because it is likely that we want to interact with the inputs.


R/ui.R
```diff
@@ -285,12 +285,14 @@ inject_remove_button <- function(x, ...) {
 #' inserting a new block within a stack.
 #'
 #' @param ns Parent namespace.
+#' @param .hidden Whether to initialise the block with
+#' hidden inputs.
 #'
 #' @export
 #' @rdname generate_ui
-inject_remove_button.block <- function(x, ns, ...) {
+inject_remove_button.block <- function(x, ns, .hidden = TRUE, ...) {
```
DivadNojnarg, line 293: thanks for seeing this. It got dropped while converting this to S3.

PR #196, opened 2024-01-17
ignore init false

christophsax:
This seems better to me, and I don't see problems so far.


DivadNojnarg:
I am good with this one @christophsax.

christophsax:
Ok, I reverted '[to correct initialization](https://github.com/blockr-org/blockr/pull/196/commits/b605740670e8645104b5dfbeb09912804d6a8d0b), so we can discuss and merge this in the separate PR.

Issue #163, opened 2024-01-09
Validation should in no way depend on field names


nbenn:
With the current validation setup, we have exclusion criteria that are hard-coded to field names. These are completely arbitrary and this logic in

https://github.com/blockr-org/blockr/blob/5d2cfcdee04b40df0a037c8bacdcb8bcd8f44897/R/utils.R#L305

should not depend on this. I just created a block where I was using a hidden field that was not called `expression` but `expr_parsed` and for a block like this, the submit button will be permanently disabled. Happy to create a reprex, but I don't think it's necessary.

christophsax:
Reprex in the linked PR #203

DivadNojnarg:
Fixed in 0.0.2

PR #203, opened 2024-01-18
Options in keyvalue field

christophsax:
Cleaned up version of an earlier, abandoned PR, #189.

It reworks the keyvalue_field and has no effects on anything else. It should be safe to merge. 

### Improved key_value field

- Now uses updateUI, instead of addUI etc.
- Respects initialization value.
- Uses colnames from data and makes them available in autocomplete
- Should be generally much more stable

### Nicolas example from previous PR

@nbenn's example from an older PR work too, it's just that value should be `NULL` rather than an empty list. The options work as expected. Having a submit button in two fields is usually not recommended.

```r

  handle_error <- function(e) {
    warning(e)
    expression()
  }

  parse_one <- function(x) {
    tryCatch(
      parse(text = x),
      error = handle_error
    )
  }

    res <- lapply(x, parse_one)

    if (length(res)) {
      res <- do.call(c, res)
    }

    res
  }

    parse_multiple(new_cols)
  }

    names(res) <- NULL
    res
  }

  suggest_by <- function(data) colnames(data)[1L]
  choices <- function(data) colnames(data)

  fields <- list(
    new_cols = new_keyvalue_field(value = NULL, submit = FALSE),
    group_by = new_keyvalue_field(value = NULL, submit = FALSE, multiple = FALSE, key = "none"),
  )

  expr <- quote(
  )

  new_block(
    fields = fields,
    expr = expr,
    ...,
  )
}

}

  do.call(
    bquote,
    list(
      attr(x, "expr"),
      where = lapply(x, value),
      splice = TRUE
    )
  )
}

    fun <- x[[cmp]]
  }

  x
}

}

```

The code above results in:

image

The two warnings are due to the fact that validation excludes `hidden_fields` based on its name and unrelated to this PR (https://github.com/blockr-org/blockr/issues/163).


christophsax:
@DivadNojnarg this should be safe to merge and would allow me to PR an admiral block with a bit of documentation.

nbenn (approved):
As discussed, ok for now.

PR #204, opened 2024-01-18
BLOCKR_DEV option fix

### LINKED ISSUE #202: `options("BLOCKR_DEV" = TRUE)` no longer working?
When doing something like that, I used to get expanded blocks, which was very useful for development. 
```r
library(blockr)
library(blockr.data)

stack <- new_stack(
  data_block,
  filter_block
)

options("BLOCKR_DEV" = TRUE)
serve_stack(stack)
```

This doesn't seem to work anymore, instead, I get:

image

JohnCoene:
This fixes #202 

Also a tiny refactor.

### DIFF
--- a/R/ui.R
+++ b/R/ui.R
@@ -182,10 +182,8 @@ generate_ui.block <- function(x, id, ..., .hidden = !getOption("BLOCKR_DEV", FAL
 
   block_class <- "block"
   inputs_hidden <- ""
-  hidden_class <- ""
   if (.hidden) {
-    hidden_class <- "d-none"
-    inputs_hidden <- hidden_class
+    inputs_hidden <- "d-none"
     block_class <- sprintf("%s d-none", block_class)
   }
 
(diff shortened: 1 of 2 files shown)


christophsax:
Nice, thanks!

DivadNojnarg (commented):
Thanks for checking. This is what I am seeing in incognito mode running @christophsax 's example:

- The stack opens with all block uncollapsed -> GOOD.
- When I collapse the stack, it only hides the last block. I would expect all the blocks to be hidden. 
- Then the toggle icons seems to be in the wrong state.

https://github.com/blockr-org/blockr/assets/18291543/cc7f2236-bc12-466b-9805-680688bec5d6

JohnCoene:
Yes but it doesn't have anything to do with `BLOCKR_DEV`.

Some recent merges, refactor, and rework of validation have thrown off much of the logic. Happy to fix but in another PR.

Issue #6, opened 2023-09-14
Code consistency issues


DivadNojnarg:
@nbenn and @christophsax :

1. I created 2 block, one using dplyr and another using ggplot2. ggplot2 uses `+` or `%+%` which does not play well with our `evalute_block.transform_block` method supporting `%>%`. I had to create a specific one which properly passes the data to the ggplot call. I have the feeling we might end up with many edge cases like this to handle. 

1. I am still trying to figure out how not to make the `generate_server.block` method growing too much by having specific block methods. This is not obvious because at some point, all blocks moreless need the same things with few exceptions: plot block won't require to display data, while other will. 

1. To switch between the default expression (identity()) vs valid expression (dplyr::select(...)), I had to tweak `generate_code.block`. The issue is that each block has its own fields, which in turn have their own rules. We end up with an endless if/else chain. This is not a good long term solution.

To be continued ...

christophsax:
2 and 3. are solved. 1 needs to be seen.

PR #209, opened 2024-01-18
start layer by layer plot vignette

DivadNojnarg:
It's currently strange to render the output of each plot block, particularly the first ggplot call which is nothing more than an empty plot. We should be able to decide whether a block renders its output. Maybe starting from the first `ggplot` block, we could go over the list of blocks within the stack that inherit from `plot_block` and if yes, we don't render the output, until we reach the last block which is allowed to render. This must be done within the stack.

TO DO:

- [x] Have nice expressions from the plot blocks.
- [x] Render only last plot block.
- [x] Fix broken tests.


DivadNojnarg:
I had to introduce a new `plot_layer_block` class with a specific `evaluate_block` method:

```r
evaluate_block.plot_layer_block <- function(x, data, ...) {
  stopifnot(...length() == 0L)
  eval(
    substitute(data + expr, list(expr = generate_code(x))),
    list(data = data)
  )
}
```

So we can use `+` instead of `%>%`. In an earlier discussion, we also [discussed](https://github.com/blockr-org/blockr/issues/6#issuecomment-1731424118) about having a further layer of abstraction such as:

```r
block_combiner <- function(x) UseMethod("block_combiner", x)

block_combiner.transform_block <- function(x) function(x, y) `%>%`(x, y)

block_combiner.plot_block <- function(x) function(x, y) `+`(x, y)

evalute_block.block <- function(x, prev, ...) {

  eval(
    substitute(fun(prev, expr), list(fun = block_combiner(x), expr = generate_code(x))),
    list(prev = prev)
  )
}
```

I did not manage to modify `generate_code.stack` such that it can elegantly combine `+` and `%>%`. TO DO: binary substitute must know how to behave with the `y` expression when it comes from a plot_layer_block or from an other block. Currently `x` and `y` are expressions which makes it impossible to use S3, we would have to pass the entire block to be able to handle this in a better way.

```r
generate_code.stack <- function(x) {

  binary_substitute <- function(x, y) {
    # TO DO: remove this hell asap (just for testing)
    if (grepl("(geom)|(labs)|(theme)", deparse(y), perl = TRUE)) {
      substitute(x + y, list(x = x, y = y))
    } else {
      substitute(x %>% y, list(x = x, y = y))
    }
  }
  Reduce(binary_substitute, lapply(x, generate_code))
}
```

DivadNojnarg:
A note for @nbenn: 

To be able to generate a valid stack expression with plot blocks, I had to introduce an early idea of yours that is the `block_combiner` and modify the `generate_code.stack` so that `Reduce` takes the list of blocks to be able to use the S3 methods and call the right `block_combiner`. The trick is that we check the class of the right element that's why we do `block_combiner(x = y, y = x)` (we apply `+` if it is a plot layer or `%>%` if not). Since `Reduce` combines elements, we also have to create a `generate_code.call` method which just returns `identity` (This is because we now pass the block elements to Reduce and not just the code expressions):

```r
#' @rdname new_stack
#' @export
generate_code.stack <- function(x) {

  binary_substitute <- function(x, y) {
    block_combiner(x = y, y = x)
  }

  Reduce(binary_substitute, lapply(x, \(b) b))
}

#' @rdname new_stack
#' @export
block_combiner <- function(x, ...) UseMethod("block_combiner", x)

#' @rdname new_stack
#' @export
block_combiner.transform_block <- function(x, y, ...) {
  substitute(
    y %>% x,
    list(x = generate_code(x), y = generate_code(y))
  )
}

#' @rdname new_stack
#' @export
block_combiner.data_block <- block_combiner.transform_block

#' @rdname new_stack
#' @export
block_combiner.plot_block <- block_combiner.transform_block

#' @rdname new_stack
#' @export
block_combiner.plot_layer_block <- function(x, y, ...) {
  substitute(
    y + x,
    list(x = generate_code(x), y = generate_code(y))
  )
}
```

This way we can get something like:

```r
get("airquality", envir = ...) %>% 
  ggplot(mapping = aes(x = .data[["Ozone"]], y = .data[["Ozone"]])) + 
  geom_point(color = "blue")
```

DivadNojnarg:
Screenshot 2024-01-19 at 10 59 11

DivadNojnarg:
My recent change to the `generate_code` breaks:

```r
code <- generate_code(
    new_stack(new_data_block)
 )
```

DivadNojnarg:
As discussed with @nbenn:

- We may not need to remove the output from intermediate layers. Currently since we only show the latest block, it could be just fine. That would indeed save us from few extra lines of code.

nbenn:
@DivadNojnarg from what you showed me today, I had a quick look at getting rid of the expression field in your example. What I think this should look like is something as

```r
new_ggplot_block <- function(data, ...) {

  new_block(
    fields = list(
      x = new_select_field(colnames(data)[1], data_cols, type = "name"),
      y = new_select_field(colnames(data)[2], data_cols, type = "name")
    ),
    expr = quote(
      ggplot(mapping = aes(x = .(x), y = .(y)))
    ),
    ...
  )
}
```

1. the select options are updated dynamically if the input data changes
2. there's no need for the expression field (which is being used far more often throughout the code base than I had envisioned this)
3. for didactic reasons, there's less indirection by using the field constructors directly

Now there's a catch: with the current code base, this does not quite work. There's a problem with a change to `type_trans()` around https://github.com/blockr-org/blockr/commit/f8f49dbe8e39f9a58647bc6a6f6e265d24dadf0e.

For the mechanics that I originally intended, we'd have to fix that quickly and then we could make the example a bit nicer I feel. What do you think?

DivadNojnarg:
@nbenn, yes such a code would be ideal in term of complexity. It seems that reverting `type_trans` to the original would work, so I have to see what else we have to change (I guess this is related to the arrange and group_by blocks which use `convert_block`. If we get rid of the latter, maybe this can work).

DivadNojnarg:
We could write `type_trans` as:

```r
type_trans <- function(x) {
  res <- value(x)

  switch(attr(x, "type"),
    literal = res,
    name = if (length(res) <= 1) {
      as.name(res)
    } else {
      unlst(lapply(res, as.name))
    }
  )
}
```

The very first version was only `as.name(res)` when type is name. However, for multi columns blocks like `group_by(col1, col2)` and `arrange` which require to use a symbol (contrary to `select(c("col1", "col2")`), we can't do `as.name(res)` which only returns the first element.

 I'll also get rid of `convert_block` to rewrite ``group_by` and `arrange` as standard blocks. Let me know if that sounds good to you @nbenn.

R/block.R
```diff
@@ -111,6 +117,12 @@ generate_code.transform_block <- function(x) {
   NextMethod()
 }
 
+#' @rdname new_block
+#' @export
+generate_code.call <- function(x) {
```
DivadNojnarg, line 122: This is due to the changes in `generate_code.stack`. After the first `Reduce` combination, `x` becomes a call (result of the 2 block codes of the previous iteration) so we don't need to render the code of something that is already code.

DivadNojnarg:
I cleaned up the show_output mechanism as discussed. This is good for review @nbenn .

R/block.R
```diff
@@ -111,6 +113,12 @@ generate_code.transform_block <- function(x) {
...
+  identity(x)
```
nbenn, line 119: Why not just return `x`?

R/block.R
```diff
@@ -138,9 +146,17 @@ evaluate_block.transform_block <- function(x, data, ...) {
 #' @param data Result from previous block
 #' @rdname new_block
 #' @export
-evaluate_block.plot_block <- function(x, data, ...) {
+evaluate_block.plot_block <- evaluate_block.transform_block
+
+#' @param data Result from previous block
+#' @rdname new_block
+#' @export
+evaluate_block.plot_layer_block <- function(x, data, ...) {
```
nbenn, line 154: I believe this method should be part of the vignette, no? The class `plot_layer_block` is introduced there and currently is not part of blockr. It might be a bit of effort to explain what this does and why it is necessary, at the same time it nicely showcases how flexible blockr is. In the case someone wants to add a block where they need yet another way of combining blocks, this'd be the way to do it w/o requiring a core change.

R/server.R
```diff
@@ -132,17 +132,19 @@ generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "pl
...
-        blk_no_srv <- blk()
-        blk_no_srv[is_srv] <- NULL    # to keep class etc
-        validate_inputs(blk_no_srv, is_valid, session)  # FIXME should not rely on input$
-
-        # Block will have a red border if any nested input is invalid
-        # since blocks can be collapsed and people won't see the input
-        # elements.
-        validate_block(blk(), is_valid, session)
-      })
+      if (display != "plot") {
```
nbenn, line 135: We'll have to revisit this sooner or later, so good for now.

R/stack.R
```diff
@@ -44,12 +57,54 @@ is_stack <- function(x) {
...
 #' @export
 generate_code.stack <- function(x) {
+  if (length(x) == 0) return(quote(identity()))
 
-  binary_substitute <- function(x, y) {
-    substitute(x %>% y, list(x = x, y = y))
+  # Handles monoblock stacks
+  if (length(x) > 1) {
+    aggregate_code <- function(x, y) {
+      block_combiner(x = y, y = x)
```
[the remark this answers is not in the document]
nbenn: This confuses me a bit. See comment for the `block_combiner()` generic.

R/stack.R
```diff
@@ -44,12 +44,54 @@ is_stack <- function(x) {
...
+
+#' Combine 2 block expressions
+#'
+#' Useful for \link{generate_code}.
+#'
+#' @rdname block_combiner
+#' @param x Block object.
+#' @param ... For generic consistency.
+#' @export
+block_combiner <- function(x, ...) UseMethod("block_combiner", x)
```
nbenn, line 68: I find the `x = y, y = x` above a bit confusing. Some thoughts:

- If dispatch should be on the second arg, couldn't this be done as

    ```r
    block_combiner <- function(x, y, ...) UseMethod("block_combiner", y)
    ```

- Maybe renaming the arugments to `curr`, `next` or so could be helpful?

R/stack.R
```diff
@@ -44,12 +44,54 @@ is_stack <- function(x) {
...
+block_combiner.transform_block <- function(x, y, ...) {
+  substitute(
+    y %>% x,
+    list(x = generate_code(x), y = generate_code(y))
+  )
+}
+
+#' @rdname block_combiner
+#' @export
+block_combiner.data_block <- block_combiner.transform_block
```
nbenn, line 82: Continuing from above, if dispatch is on the second block, this will never be used, no?

R/stack.R
```diff
@@ -44,12 +44,54 @@ is_stack <- function(x) {
...
 
-  Reduce(binary_substitute, lapply(x, generate_code))
+#' @rdname block_combiner
+#' @export
+block_combiner.plot_block <- block_combiner.transform_block
+
+#' @rdname block_combiner
+#' @export
+block_combiner.plot_layer_block <- function(x, y, ...) {
```
nbenn, line 90: See comment above about this being part of the vignette. If you feel that this is out of scope for the vignette, it could maybe go into an "advanced" section at the end? The reason why I feel it should go there is two-fold:

- code org: as long as we don't have constructors for these objects in blockr, blockr should not contain implementations for relevant generics
- didactic aspects: this is how one should go about adding such functionality in case they do not have write privileges to the core repo

If you ran into trouble registering the generic implementations in the vignette, maybe `.S3method()` helps? In case you are unfamiliar see https://github.com/blockr-org/blockr/pull/189#issuecomment-1891936441 for an example.

nbenn (approved):
Looks good. The only thing I think we should change is moving the `block_combiner()` and `evaluate_block()` implementations for `plot_layer_block` into the vignette (see comments on the reasoning behind that). Other comments are optional.

R/block.R
DivadNojnarg: LOL :)

R/stack.R
DivadNojnarg: `next` is reserved but we could indeed change the arguments for better clarity.

nbenn: Of course it is.

DivadNojnarg: But i am fine with `left` and `right`.

R/stack.R
DivadNojnarg: Yes I just removed it.

PR #210, opened 2024-01-19
no-validation-for-plots

### LINKED ISSUE #205: Validation also applied to plot and data blocks
Initially, the validation was only applied to `transform_block` since it is too experimental. In the latest version with the new `generate_server_block` function, it is applied to all blocks. We can restore the old behavior with:

```r
# Validate block inputs
      if (is.null(display)) {
        obs$validate_inputs <- observeEvent(r_values(), {
          message(sprintf("Validating block %s", class(x)[[1]]))
          blk_no_srv <- blk()
          blk_no_srv[is_srv] <- NULL    # to keep class etc
          validate_inputs(blk_no_srv, is_valid, session)  # FIXME should not rely on input$

          # Block will have a red border if any nested input is invalid
          # since blocks can be collapsed and people won't see the input
          # elements.
          validate_block(blk(), is_valid, session)
        })
      }
```

christophsax:
As suggested by @DivadNojnarg, fixes #205

### DIFF
--- a/R/server.R
+++ b/R/server.R
@@ -132,17 +132,19 @@ generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "pl
         bindEvent(r_values(), in_dat())
 
       # Validate block inputs
-      obs$validate_inputs <- observeEvent(r_values(), {
-        message(sprintf("Validating block %s", class(x)[[1]]))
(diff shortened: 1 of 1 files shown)


DivadNojnarg (approved):
Thanks.

PR #212, opened 2024-01-23
Fix: Rendering error block and stack collapse state

### LINKED ISSUE #206: Block and Stack collapse
Fix collapse issues described in https://github.com/blockr-org/blockr/pull/204

Also fix issue with validation input border (valid input keeps red border, should be green)

JohnCoene:
This initially intended to fix #206 
But I had not realised that problem reported in #206 is solely caused by the `BLOCKR_DEV` option, everything works fine without this.

This instead 1) improves the error handling on blocks, namely it forces the expansion of a stack if it any of its blocks report an error, 2) fixes the initial state of an expanded stacks (when initialised empty).


JohnCoene:
This can be reviewed/merged

DivadNojnarg (commented):
There is a tiny inconsistency in the state of the last block collapse icon. See below (I toggle the stack and the last block. At some point, I end up with a wrong icon state).
https://github.com/blockr-org/blockr/assets/18291543/645e4aea-6c9b-491a-a9d8-08df574b5e6a

JohnCoene:
Fixed.

Some of the logic for this changed when we refactored/changed how the output and code UI work. It used to be that these buttons were placed in the "block title" which is hidden when only the last output of a stack is shown.

Now that those buttons are outside of the title they no longer were hidden.

It does not make sense/breaks the logic to be able to expand/collapse when the stack is not in "edit" mode.

This is fixed now.

DivadNojnarg (commented):
> Now that those buttons are outside of the title they no longer were hidden.

Yes I can confirm I don't see the block level collapse button when the stack is collapsed. 

I still end up in an inconsistent state where the last block toggle does not reflect the correct opening state after some UI interactions. To reproduce: 

1. Expand the stack
2. Collapse the last block
3. Collapse the stack.
4. Expand the stack. 

Screenshot 2024-01-26 at 16 52 52

DivadNojnarg:
@JohnCoene: do you manage to reproduce the last point I raised?

DivadNojnarg (commented):
Thanks @JohnCoene for looking into this. I still see weird state. 

```r
pkgload::load_all()

stack <- new_stack(
    data_block,
    select_block
)

serve_stack(stack)
```

Below is what happens when I run the app and open the stack: 
- The first block collapse icon is in an incorrect state (compared to the select block).
Screenshot 2024-01-31 at 19 06 59

- This persists when I collapse the first block (see below).

Screenshot 2024-01-31 at 19 10 29

- If I recollapse the stack, I can re-sync toggle icons to be consistent.

JohnCoene:
@DivadNojnarg this should be good to merge now

DivadNojnarg (approved):

JohnCoene:
Thank you David!

PR #222, opened 2024-01-26
Small ui tweaks

JohnCoene:
This implements small UI tweaks to attempt to make the UI slightly more compact.

- Reduce margin and padding in places
- Alignment consistency
- Remove `cube` and `cubes` icon on stack
- Reduce height of stack and block headers

image

Fore reference, how it currently looks.

image

Small but noticeable changes


DivadNojnarg (commented):
That's not a big thing, but I was thinking that for the end user, icons could be a good way (at least better than the block name which is raw text) to convey block information details, for instance whether it is a data, transform or plot block. What do you think? Currently I was using the same icon for all block types but this can change depending on the block class, that is have a generic and methods to create the block metadata icon.

JohnCoene:
I like the idea of communicating the block type to the user.

It's just that currently we use the same icons everywhere so it doesn't say anything and makes the UI needlessly busy.

We could use icons for that, with tooltip. Sounds like a good idea.

DivadNojnarg:
> It's just that currently we use the same icons everywhere so it doesn't say anything and makes the UI needlessly busy.

Yes that's true. That's why I am happy if we can change these with relevant icons and our S3 functions (`block_icon`, `block_icon.data`, `block_icon.plot`). I can also do the change if you like.

JohnCoene:
Let me open an issue and tackle this in this PR, switching back to draft

JohnCoene:
@DivadNojnarg added icons and tooltip based on block type

image

DivadNojnarg (approved):
Thanks for this improvement. LGTM

PR #231, opened 2024-01-29
Feat: data parser blocks

DivadNojnarg:


R/block.R
```diff
@@ -22,11 +22,15 @@ new_block <- function(fields, expr, name = rand_names(), ...,
                       class = character(),
                       layout = default_layout_fields) {
   stopifnot(
-    is.list(fields), length(fields) >= 1L, all(lgl_ply(fields, is_field)),
+    is.list(fields),
     is.language(expr),
     is_string(name)
   )
 
+  if (length(fields) > 0) {
```
DivadNojnarg, line 30: Since the data parser block has 0 fields... See reason in `new_csv_parser_block`

R/block.R
```diff
@@ -124,6 +128,17 @@ evaluate_block.data_block <- function(x, ...) {
   eval(generate_code(x), new.env())
 }
 
+#' @rdname new_block
+#' @export
+evaluate_block.csv_parser_block <- function(x, data, ...) {
```
DivadNojnarg, line 133: TO DO: this won't work if this is coming from the filebrowser block. TO improve.

R/block.R
```diff
@@ -260,6 +261,24 @@ filesbrowser_block <- function(...) {
   initialize_block(new_filesbrowser_block(...))
 }
 
+#' @rdname new_block
+#' @export
+new_csv_parser_block <- function(data, ...) {
+  new_block(
+    # Empty list
+    fields = list(),
```
DivadNojnarg, line 269: I initially had a string field which was disabled and containing the data path from the previous block. However, the string field fails to updated when I changed the upstream data path...

DivadNojnarg:
Another approach (removed `server_code.upload_block`, `evaluate_block.csv_parser_block` and added `generate_code.upload_block`):

```r
pkgload::load_all()

generate_code.upload_block <- function(x) {
  eval(generate_code.block(x), new.env())
}

new_upload_block <- function(...) {

  get_path <- function(dat) {
    if (length(dat) == 0) return(character())
    dat$datapath
  }

  new_block(
    fields = list(
      dat = new_upload_field()
    ),
    expr = substitute(fun(.(dat)), list(fun = get_path)),
    ...,
    class = c("upload_block", "data_block")
  )
}

upload_block <- function(...) {
  initialize_block(new_upload_block(...))
}

new_csv_parser_block <- function(data, ...) {

  read_data <- function(data) {
    if (length(data) == 0) return(quote(identity))
    quote(utils::read.csv())
  }

  new_block(
    fields = list(
      expression = new_hidden_field(read_data)
    ),
    expr = quote(.(expression)),
    ...,
    class = c("csv_parser_block", "transform_block")
  )
}

csv_parser_block <- function(data, ...) {
  initialize_block(new_csv_parser_block(data, ...), data)
}

stack <- new_stack(upload_block, csv_parser_block)
serve_stack(stack)
```

R/server.R
```diff
@@ -162,7 +162,7 @@ generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "pl
       } else {
         reactive({
           req(is_valid$block)
-          if (is.null(in_dat())) {
+          if (is.null(in_dat()) && !inherits(x, "transform_block")) {
```
DivadNojnarg, line 165: @nbenn I did not notice but the block server refactor introduced an issue. At this line, it is possible to have evaluate a transform block without data parameter, leading to an error when calling `evaluate_block` (`evaluate_block.transform_block` expects a data parameter). `generate_code_server` is not a generic so we lost the ability to tackle this is in a clean way. I have to add `!inherits(x, "transform_block"` to continue my PR but we will have the same issue with plot blocks I think.

R/block.R
DivadNojnarg: I reverted this in the latest commits since I am using the `new_hidden_field` again. That said, we could maybe think about allowing block with empty fields.

R/block.R
DivadNojnarg: Also resolved in latest commits. This function was dropped and I use the default one.

DivadNojnarg:
@nbenn, as discussed, the generated stack expression is always wrong and it actually seems that the existing upload_block does not generate proper expression. This is independent from the recent changes to the stack code generation (I tried before merging main into that branch, and the problem is the same).

To reproduce, you need a csv file and:

```r
pkgload::load_all()

stack <- new_stack(
    upload_block,
    csv_parser_block,
    select_block
)

serve_stack(stack)
```

You can upload your csv, see that the parser contains the data and you can select columns. However, copying the stack code gives `NULL %>% identity() %>% identity()`. I guess this might be because the upload_field does not have initial values and we also don't have any `updateFileInput` on the shiny side, so update_field.upload_field does nothing.

R/block.R
```diff
@@ -209,20 +221,16 @@ data_block <- function(...) {
...
-      .(path),
-      list(path = dat$datapath)
-    )
+  data_path <- function(file) {
+    if (length(file)) file$datapath else character()
   }
 
   new_block(
     fields = list(
-      dat = new_upload_field(),
```
nbenn, line 224: @DivadNojnarg Not entirely sure but the problem here might be partial argument matching as `dat` is a substring of `data`.

R/block.R
```diff
@@ -209,20 +221,16 @@ data_block <- function(...) {
...
+      file = new_upload_field(),
       expression = new_hidden_field(data_path)
     ),
-    expr = quote(.(expression)),
```
nbenn, line 227: @DivadNojnarg this I believe does not work as intended. If you wrap your string in `c()` all is fine.

nbenn:
@DivadNojnarg I'll quickly go over the parser issues as well.

DivadNojnarg:
@nbenn: what I initially wanted for the parser is to have a string field that contains the path coming from the previous upload block (in case the previous block is collapsed, user may want to see the path). We'd have to make this field disabled to avoid issues when user accidentally change it (possible with Bootstrap but we'd need to modify the string field to add the disabled attribute (shiny::textInput does not accept ... to pass any exta attribute).

nbenn:
@DivadNojnarg playing around with this a bit more, what we have, in principle, I believe should work -- if it were not for the issue that we're trying to solve in #234. The `input$copy` action grabs the content of `vals$stack`, which is not properly updated.

https://github.com/blockr-org/blockr/blob/42f6cb06a2b0a3fee90aa212a16df4f05bb2fa28/R/server.R#L317-L327

DivadNojnarg:
Thanks for these insights, that makes sense. So let's finish #234 :)

DivadNojnarg:
@nbenn. All parsers are now working with the latest changes in main including the stack copy code. 

In the long run, we might want to implicitly call the suitable parser block right after a data upload/browser block to avoid user to have to explicitly pass them in the stack. For Friday's demo, I won't add more since we need to focus on other issues.

nbenn (approved):
Looks good I think. At some point we might want to expose some args for the parsers and maybe move them to another pkg. But for now, this is good!

Issue #112, opened 2023-12-01
Stack/block names/IDs


nbenn:
Are stack and block names/IDs part of the respective objects? If yes, who is responsible for ensuring uniqueness?

DivadNojnarg:
Following #119:

- It would be easier to have stack holding all the blocks IDs. 
- We should follow the same principle for the workspace where stack IDs would be stored.

PR #234, opened 2024-01-30
Properly update workspace

### LINKED ISSUE #226: Workspace updating
This is split off of #221 and concerns

> I think most of the workspace handling could be internalised within {blockr}, we need to update the workspace when a block is added, removed to a stack or when a block is updated. All of this can be observed within blockr but I currently have to handle that outside of blockr.

nbenn:
Fixes #226.


nbenn:
Continuing from https://github.com/blockr-org/blockr/issues/226#issuecomment-1916543949. Further observations:

- The workspace issue also holds for stacks: we are not properly keeping the stack object up to date.
- We are lacking clarity on what the 3 server functions for block, stack and workspace should return, i.e. what objects we need available from a lower level in a higher one (i.e. do we need the reactive version of a block available in the stack to trigger updates?)

nbenn:
@DivadNojnarg with the current state([d16c897](https://github.com/blockr-org/blockr/pull/234/commits/d16c8974a86807d89c08add3f1c66368e0fe9a81)), I get for a setup as 

```r
devtools::load_all()
serve_workspace()
```

and UI actions

1. Add stack
2. Add new block
3. Choose dataset_block

an empty stack with no dataset block. I'm still trying to figure out why. Maybe you have an idea?

R/server.R
```diff
@@ -486,14 +494,23 @@ generate_server.workspace <- function(x, id = NULL, ...) {
...
           el,
-          id = attr(el, "name"),
-          new_block = reactive(vals$new_block[[attr(el, "name")]])
+          id = stack_id,
+          new_block = reactive(vals$new_block[[stack_id]])
         )
 
         # Handle new block injection
-        inject_block(input, vals, id = attr(el, "name"))
+        inject_block(input, vals, id = stack_id)
```
nbenn, line 504: I'm messing with IDs here, so this might cause problems with how input elements that are listened to in `inject_block()` are named.

nbenn: But no, `stack_id == attr(el, "name")` I believe.

nbenn: Do we have a namespacing issue? Should IDs in `inject_block()` be constructed with `myworkspace-` (or whatever the workspace ID is) prepended?

R/server.R
```diff
@@ -525,21 +542,24 @@ init <- function(x, ...) {
...
 
   observeEvent(TRUE, {
-    lapply(seq_along(stacks), \(i) {
-      id <- attr(stacks[[i]], "name")
-      vals$stacks[[i]] <- generate_server(
-        stacks[[i]],
-        id = id,
-        new_block = reactive(vals$new_block[[id]])
+
+    for (nme in names(stacks)) {
```
nbenn, line 550: One thing I'm trying to do here is consistently refer to stacks by their name/ID and not by position. I'm sorry if this is causing issues. Happy to elaborate on my reasoning behind this.

R/server.R
DivadNojnarg: We had to create the stack/blocks remove buttons from the parent namespace to avoid back and forth between modules.

DivadNojnarg:
When adding a block, I get a JS warning (Chrome):

```
The selector you chose ("#myworkspace-lqxloprtieehasv-lqxloprtieehasv-body") could not be found in the DOM.
```

That's a namespace issue, as you see, the stack namespace appears twice. Now I need to find where is the issue.

DivadNojnarg:
A separate note:

- There is an issue to remove stacks, which works well except for the last one. This is not handled R side with `removeUI` but client side and not consistent with what we have for the blocks (removed R side). cc @JohnCoene.

R/server.R
DivadNojnarg: I think this change won't cause issues. See below. We could do the same comment at the stack server function level, where blocks are handled by position in `vals$blocks`. That could also be updated (maybe in another PR).

R/server.R
```diff
@@ -297,8 +297,8 @@ generate_server.stack <- function(x, id = NULL, new_block = NULL, ...) {
           } else {
             insertUI(
               selector = sprintf(
-                "#%s-body",
-                session$ns(attr(vals$stack, "name"))
+                "#%sbody",
+                session$ns("")
```
nbenn, line 301: @DivadNojnarg many thanks for your help on this!

nbenn:
@DivadNojnarg the issue you describe above (`could not find function "in_dat"`) boils down to how `vals$blocks[[i - 1]]$data` in `init_block()` is not a reactive, but the value itself.

What confuses me is that

```r
Browse[1]> vals$blocks[[i - 1]]
<ReactiveValues>
  Values:    block, data
  Readonly:  FALSE
```

so I don't understand why `vals$blocks[[i - 1]]$data` does not error when evaluated in a non reactive context.

DivadNojnarg:
> so I don't understand why vals$blocks[[i - 1]]$data does not error when evaluated in a non reactive context.

Maybe because `init_block` is called by `init_blocks`, the latter being in an observe (so a reactive context):

```r
observeEvent(TRUE, {
    for (i in seq_along(x)) {
      vals$blocks[[i]] <- init_block(i, vals, session)
    }
  })
```

nbenn:
@DivadNojnarg the current issue is as follows:

1. startup as `serve_workspace()`
2. add stack, reports `ADD STACK (stack_id)`
3. add `dataset_block`, reports `ADD BLOCK (position 1)`
4. add `filter_block`, reports `ADD BLOCK (position 1)` which indicates that something is wrong, also 2nd block is not shown
5. add `select_block`, reports `ADD BLOCK (position 2)` and crashes app

DivadNojnarg:
@nbenn. Is the purpose of the following event to update the reactive version of the stack with any block change (say if I change the data choice in a data block)?

```r
observeEvent(vals$blocks, {
        if (length(vals$blocks)) {
          browser()
          blks <- lapply(vals$blocks, `[[`, "block")
          vals$stack <- set_stack_blocks(vals$stack, blks)
        }
      })
```

`vals$blocks` is actually a wrong trigger as I will only trigger when the length of this list changes (but not when any of its element change: each block returns data and the block element). Instead we want something like:

```r
lapply(vals$blocks, \(b) {
          list(
            b$data(),
            b$block()
          )
        })
```

Also, this event will fire right after adding a block, we is not necessary because `add_block` already calls `set_stack_blocks` (this is not harmful but not optimal). I am going to push some changes to reflect this. 

This seems to fix our issue.

DivadNojnarg:
@nbenn The change made at the stack level should also be propagated at the workspace level when we listen to (in generate_server.workspace):

```r
observeEvent(vals$stacks, {
        if (length(vals$stacks)) {
          set_workspace_stacks(
            lapply(vals$stacks, `[[`, "stack"),
            force = TRUE
          )
        }
      })
```

The trigger should be something like (untested):

```r
lapply(vals$stacks, \(s) {
          list(
            s$stack(),
            s$blocks()
          )
        })
```

R/server.R
```diff
@@ -331,9 +336,18 @@ generate_server.stack <- function(x, id = NULL, new_block = NULL, ...) {
         set_title(vals$stack, input$newTitle)
       })
 
-      observeEvent(vals$blocks, {
+      # Any block change: data or input should be sent
+      # up to the stack so we can properly serialise.
+      observeEvent({
+        lapply(vals$blocks, \(b) {
```
nbenn, line 342: @DivadNojnarg many thanks for your insight here!

R/server.R
```diff
@@ -512,13 +505,11 @@ generate_server.workspace <- function(x, id = NULL, ...) {
         inject_block(input, vals, id = stack_id)
       })
 
-      observeEvent(vals$stacks, {
```
nbenn: @DivadNojnarg how about this for a trigger? No need to create a call as this is a `reactivevalues` object, so simply accessing the element should return the current value (IIUC).

nbenn:
@DivadNojnarg we're making progress. We still don't have everything set up correctly though. One issue is that we currently don't have "dynamic" routing of block results.

What I mean by this is if we for example remove a block in the middle of a stack (position i), then before removal, the output of i was the input of i + 1. After removal we have to update i + 1 such that the result of i - 1 becomes its input.

A reprex of the problem is

1. initialize as `serve_workspace()`
2. add a dataset block
3. add a filter block and set values (e.g. `Ozone > 20`
4. add a select block
5. remove filter block

This will leave you with improperly subsetted data.

(I tried to reproduce this issue without the changes in this branch to see whether that is something that I/we introduced here. Unfortunately I could not. Current main, as well as older versions of main such as a487403f18f764e57db1c8a444a5da3b94e59b31 or 54c027d415fa576a2992d259f07c3fc294acb698 all suffer from other issues.

```r
devtools::load_all()

ws <- get_workspace()

shinyApp(
  ui = bslib::page_fluid(
    generate_ui(ws, id = "myworkspace")
  ),
    generate_server(ws, id = "myworkspace")
  }
)
```

1. add dataset block
2. add filter block

app crashes with `Error in dataset_block: could not find function "dataset_block"`.)

DivadNojnarg:
> @DivadNojnarg we're making progress. We still don't have everything set up correctly though. One issue is that we currently don't have "dynamic" routing of block results.
> 
> What I mean by this is if we for example remove a block in the middle of a stack (position i), then before removal, the output of i was the input of i + 1. After removal we have to update i + 1 such that the result of i - 1 becomes its input.
> 

Isn't it what's here or partially: https://github.com/blockr-org/blockr/blob/226-update/R/server.R#L407? We reinitialise the downstream blocks with the correct data. Unfortunately, I think we may end up with input/output observer clash because blocks have the same id, despite being at different positions.

R/server.R
nbenn: Thanks, added a log entry and it passes these checks.

DivadNojnarg:
@nbenn The block routing from `handle_remove` was never reached due to a swapped line. `vals$stack[[to_remove]] <- NULL` was exectuted before entring the if condition, meaning that the code inside was never reached. Below is what I changed and it allows us to correctly update downstream blocks when you remove a block in the middle of the stack:

```r
vals$blocks[[to_remove]] <- NULL
    # Reinitialize all the downstream stack blocks with new data ...
    if (to_remove < length(vals$stack)) {
      vals$stack[[to_remove]] <- NULL
      for (i in to_remove:length(vals$stack)) {
        attr(vals$stack[[i]], "position") <- i
        vals$blocks[[i]] <- init_block(i, vals)
      }
    }
```

Feel free to add a logging message as I feel it might cause us issues. I still don't know whether we have to reset observers/input/ouput in case of reordering of blocks ...

nbenn:
@DivadNojnarg just a quick update here: I'm currently chasing down another bug. With a setup as

```r
serve_workspace(my_stack = new_stack(data_block))
```

the stack is rendered properly, but not the data block (missing completely). The issue afaik is related to #112: currently we have a mix of referring to this stack as `my_stack` and the randomly generated ID (as `attr(., "name")`).

I'm trying to change things such that the workspace manages stack IDs/the stack manages block IDs, etc.

I also reworked the workspace API slightly such that the workspace can be passed as argument (should not be a breaking change though). I'm sorry, the amount of change that is accumulating in this PR is starting to get out of hand. Merging this we'll have to be careful that we don't break things shortly before the demo.

nbenn:
@DivadNojnarg I think we're making progress here. Sorry for the many (semi-)related changes all made in this PR.

Also sorry for some of the inconsequential (style) changes I made in some places that crowd the diff a bit. I was trying to make styling a bit more homogenous. At some point we can discuss some minor styling things to converge a bit more but for now we have more urgent things I believe (e.g. #262).

Let me know if you run into issues. I'll remove the WIP status on the PR as we probably should merge this soon. It might cause some issues as we're messing with some more fundamental parts here. I hope it's not too bad.

nbenn:
One more thing: I made some changes that the workspace manages stack names, but have not yet propagated this to stacks managing block names. As stuff seems to work now, I'd tackle this in a separate PR.

nbenn:
@DivadNojnarg One thing I couldn't quite figure out: I was trying to add a test as

```r
test_that("remove stack from workspace", {

  skip_on_cran()

  driver <- shinytest2::AppDriver$new(
    serve_workspace(my_stack = new_stack(data_block, filter_block)),
    name = "manipulate-workspace"
  )

  driver$click("myworkspace-my_stack-remove-stack")

  expect_length(get_workspace_stacks(), 0L)

  driver$stop()
})
```

But that unfortunately does not work. All is fine if I perform the click manually in the browser. Am I doing something wrong with shinytest2?

DivadNojnarg:
@nbenn . Thanks for the progress here. Regarding your latest comment about shinytest2, I gave it a quick try with your code, you can inspect the headless session with:

```r
driver$view()
```
which should open a special chrome session. From that session, I am also unable to manually click on the delete stack button. Also I noticed the workspace state seems wrong (indicated n stacks = 0) and the filter block does not appear. 

Screenshot 2024-02-03 at 22 05 16


When I run your code outside shinytest2, it works correctly.

For now, I would not bother with that. In any case, this is on my todo list when we have a stable-enough server side code. We'll have to use `exportTestValues` (https://rstudio.github.io/shinytest2/articles/robust.html#exported-values) to correctly test things.

DivadNojnarg:
On a side note, I can confirm that the stack code is now correctly updated as you mentioned in #231.

DivadNojnarg:
@JohnCoene: could you test this branch and let us know if you see major breaking elements for your application?

JohnCoene:
So if I understand correctly this branch handles the update the workspace internally, correct?

In the app we initialise stacks empty, this fails on my end.

```r
library(blockr)
library(blockr.data)
library(shiny)

stack <- new_stack()

ui <- fluidPage(
  generate_ui(stack)
)

  s <- generate_server(stack)
}

shinyApp(ui, server)
```

```
Error in generate_ui.stack(stack) :
  argument "id" is missing, with no default
```

DivadNojnarg:
I think if not controlled by the workspace, you need to pass the stack id as a parameter like here. https://github.com/blockr-org/blockr/blob/226-update/inst/examples/add-blocks/app.R

nbenn:
@JohnCoene Yes, as @DivadNojnarg points out, this is a breaking change introduced here:

It is no longer the stack that "manages" its ID, but the enclosing context (i.e. the workspace). For that reason you have to pass the ID from that context to the stack. This is mainly for guaranteeing uniqueness.

If this causes issues for you, we can revert to the previous behavior where the stack-managed ID is used if no value is passed from the parent context. But this would be more of a temporary workaround.

A similar change will be coming to blocks, where it is the stack that should manage block ID and position, but in a separate PR.

JohnCoene:
Apologies but I do not understand. What do we mean by 

> if not controlled by the workspace

How can I leave it controlled by the workspace? I thought that was what this PR is about: having blockr control the workspace.

It's may not be managed by blockr but it still needs to be managed; we needlessly push that workload onto blockr users. Unless there are advantages to leaving this to the user which I do not see?

For uniqueness, 1) if the workspace if managed by blockr we could check that the generated id is not already in use, 2) dramatically increase entropy on `rand_names`, 3) use a UUID generator.

Also, I'm confused about the impact it has on the serialisation/deserialisation. How I can deserialise a stack yet match it to its original ID (which I suppose I have to store somewhere else)?

Apologies but I genuinely do not understand what such a change brings to blockr, particularly so close to our all too important deadline, it would really be much easier for me if we could revert how this works.

nbenn:
@JohnCoene I think we are misunderstanding each other here. If you start up a multi-stack app as

```r
serve_workspace(stack_1 = new_stack(data_block), stack_2 = new_stack(data_block))
```

It is the workspace that checks whether stack IDs are unique. If not, this will error, as

```r
serve_workspace(stack = new_stack(data_block), stack = new_stack(data_block))
```

The code snippet you share above is not a multi-stack setting. To make that work, you just need to call as

```r
stack <- new_stack()
id <- "my_stack"

ui <- fluidPage(
  bslib::page_fluid(generate_ui(stack,  id))
)

  generate_server(stack, id)
}

shinyApp(ui, server)
```

But there currently is an issue with rendering that properly so, I'll check that out. As I said, I can bring back that the stack-internal ID can be used as default for the above setting. But that'd be a temporary workaround for now.

The more fundamental misunderstanding I believe is that for making use of the functionality we're trying to bring in here your entrypoint should be the workspace and not the stack, i.e. something like

```r
ws <- get_workspace()

ui <- bslib::page_fluid(
  generate_ui(ws, id = id)
)

  generate_server(ws, id = id)
}

shinyApp(ui, server)
```

DivadNojnarg:
After checking, the add block listener fires for stack1 and the id looks correct:

```r
observeEvent(input[[listener_id]], {
    browser()
    # Reset to avoid re-adding existing blocks to stacks
    vals$new_block <- NULL
    block <- available_blocks()[[input[[sprintf("%s-selected_block", id)]]]]
    # go at the end)
    vals$new_block[[id]] <- list(block = block)
  })
```

However, this does not propagate to the stack server function. Still need to check why.

DivadNojnarg:
@nbenn The issue is in `init.workspace`:

The use of the for loop seems to let `nme` always being equal to 2. When I used this instead with `lapply`, I manage to get it working (I can add a select block to stack 1 and then a filter block to stack2):

```r
lapply(names(stacks), \(nme) {

      handle_remove(stacks[[nme]], vals, nme, workspace = x)

      vals$stacks[[nme]] <- generate_server(
        stacks[[nme]],
        id = nme,
        new_block = reactive({
          browser()
          vals$new_block[[nme]]
        })
      )

      # To dynamically insert blocks
      inject_block(input, vals, nme)
    })
```

nbenn:
@DivadNojnarg ofc I was the one who introduced the for-loop instead of the `lapply()` construct you had in place originally. I'm sorry I don't account for how some transformations are not valid in this context.

nbenn:
@JohnCoene & @DivadNojnarg I made the changes as we discussed. Can you guys check it out and tell me if stuff does not work as expected?

@JohnCoene many thanks for the discussion earlier. I have to say, initiating these updates from the stack (instead of the workspace) seems the smarter thing to do (by a lot!).

JohnCoene:
Great, thank you so very much @nbenn !

I have tested that within the existing application without any kind of serialisation/desirialisation and it works like charm.

DivadNojnarg:
@nbenn: On my side, I tested to create few stacks, add blocks, remove blocks, clear workspace, restart from scratch, ... which seems to work well (no error and data are consistent).

nbenn:
@JohnCoene @DivadNojnarg thanks for the feedback. Fingers crossed!

nbenn:
@JohnCoene, @DivadNojnarg some thoughts/notes on the discussion we had today.

- The main reason for me to remove the id defaults to `generate_ui.stack()` and `generate_server.stack()` was that with you can do something like (note the missing id argument in the `generate_server()` call

   ```r
    id <- "my_stack"
    st <- new_stack(data_block)

    ui <- bslib::page_fluid(generate_ui(st, id))

      generate_server(st)
    }

    shinyApp(ui, server)
   ```

    You get a non-functional app with no indication as to what went wrong. For that I think we should (in the long run, not this week) either only use the stack-internal ID or the ID passed from the parent scope.

- What I meant with "the workspace should manage IDs" is the following: It is perfectly valid to create two stacks as

    ```r
    stack1 <- new_stack(data_block, name = "stack1")
    stack2 <- new_stack(data_block, name = "stack1")
    ```

    Nothing wrong with this. If the stack-internal IDs are what is actually however, it should not be possible to assign them both to the same workspace. I designed the workspace to hand out names (as `set_workspace(stack1 = stack1, stack2 = stack2)` and with the stacks from above have no name-clash) with the intention to remove the stack-internal names. But we do not necessarily have to do that. We can also check the stack-managed for uniqueness when assigning to the workspace.

    With stacks in the workspace the issue could also be solved with sufficiently long randomly generated IDs (or something like GUIDs) as a kind of workaround to have low probability of accidental clashes. A similar issue however pops up with blocks in a stack, where we have to handle both IDs *and* positions. Unlike the IDs where you can set once, positions have to be actively managed and the only place where this reasonably can happen is in the stack.

    It seemed natural for me to treat both cases the same in that the parent manages IDs of the child. **But** this does not mean that IDs cannot still be a property of the child. This is fine too and "managing" then does not include giving out the IDs but just checking uniqueness.

- An important realization (at least for me) today was (thanks to you pushing in that direction @JohnCoene) that while the parent scope is responsible for adding/removing children, it should probably not be responsible for keeping an existing child up to date. We now have this set up for workspace/stack and should propagate this to stack/block in the near future.

Issue #239, opened 2024-01-31
Arbitrary code block


JohnCoene:
It'd be good to have an arbitrary code block. I believe the idea was given before as some form of "escape block" so that one can always create the necessary block if it does not exist in the registry.

Moreover, this is greatly reduces the time from script to blockr.

I'm happy to take this one with either the shinyace package or the ace editor directly.

nbenn:
@JohnCoene I'm closing here, as I believe this was added to blockr.extra in https://github.com/blockr-org/blockr.extra/pull/7.

PR #248, opened 2024-02-01
Code block

JohnCoene:
Initial draft of code block #239 

@DivadNojnarg @nbenn inputs welcome; the idea is to have a single block for arbitrary code that can output a plot or data.frame, I'm not sure how feasible this is, two separate blocks may be better suited, thoughts?

EDIT: dynamic output (DT/plot) also implies dynamic or loose `ìnput`/`output` in the registry which sounds a bit more complicated perhaps?

image


JohnCoene:
Maybe @nbenn knows a better and safer way to evaluate the code. What I have works but it certainly feels a bit dangerous. Unsure how much we can protect from malicious expressions.

https://github.com/blockr-org/blockr/blob/e65a9629e78f5f99554652887f6843f524739f10/R/block.R#L218

nbenn:
@JohnCoene there are a couple of open questions for such an approach.

- As you point out, safety. We can of course blacklist some functions like `system()`, `system2()`, ..., but such an approach only gets you so far. We could think about evaluating the code in a sub-process that is run as another user with limited privileges. I have never done anything like that from R and this comes with all sorts of portability issues etc. I don't think it's going to be super easy to get this right.
- In an execution environment where the user is not in charge of package management (like Connect for example), they might be limited to packages that are already there. I don't know what happens when you call `install.packages()` there, whether this persists over sessions, etc.

I'm in no way trying to question the usefulness of such a block. I'm only saying that making this safe & stable might not be straightforward.

JohnCoene:
Do you think we could merge that for the demo? If OK with you I would then open a separate issue to look into "validating" the `code_field` so we can improve safety/security.

It'd be nice to show and maybe even useful as fallback in case something goes wrong with another block.

Let me know if there are things in here that should be changed.

DivadNojnarg:
@JohnCoene What would be the effort to move the block to {blockr.extra}?

Issue #247, opened 2024-02-01
Logging


JohnCoene:
We currently log a lot of things, I'm not sure it's all that useful.

E.g.: dropping a data block

```
Updating block dataset_block
Updating UI of block dataset_block
Validating block dataset_block
Updating block dataset_block
Updating UI of block dataset_block
Validating block dataset_block
Updating block dataset_block
Updating UI of block dataset_block
Validating block dataset_block
Updating block dataset_block
Updating UI of block dataset_block
```

DivadNojnarg:
It could be toggled with the BLOCKR_DEV env var.

PR #252, opened 2024-02-01
Logging

JohnCoene:
This PR implements issue described in #247: messages are logged only when he options `BLOCKR_DEV` is `TRUE`.

Happy to modify further, let me know what you think.


nbenn:
@JohnCoene I think this is a good change. I have some ideas on how the logging API could look like, I'll add here. All of this is completely optional, just take what you like.

```r
as_log_level <- function(level) {

  log_levels <- c("error", "warn", "info", "debug", "trace")

  if (is.numeric(level)) {

    stopifnot(is_count(level), level >= 1L, level <= length(log_levels))

    level <- log_levels[level]
  }

  ordered(match.arg(level, log_levels), log_levels)
}

log_level <- function() {
  as_log_level(
    getOption(
      "BLOCKR_LOG_LEVEL",
      if (getOption("BLOCKR_DEV", FALSE)) "debug" else "info"
    )
  )
}

write_log <- function(..., level = "info", emitter = default_logger()) {

  if (as_log_level(level) > log_level()) {
    return(invisible(NULL))
  }

  emitter(...)

  invisible(NULL)
}

default_logger <- function() getOption("BLOCKR_LOGGER", log_msg)

log_msg <- function(...) message(...)
```

Some thoughts

- I think it'd be good to have a single function, `write_log()` here where the level is an arg (from this we can still create wrappers like `log_info()`, `log_warn()`, etc.
- If we make the function that actually emits the logs customizable, we can also write to file (DB or whatever else) and decide what to do at runtime.
- I'm sticking to the standard set of logging levels (Log4j inspired I believe). Maybe we drop the "trace" level, but we can also keep it (doesn't hurt I guess).
- The log level is not fixed to the "DEV" setting, but can be configured independently and to any of the levels (i.e. you could do `options(BLOCKR_LOG_LEVEL = "warn")` and you would only get errors and warnings.

nbenn:
@JohnCoene regarding my comments: they were just some suggestions. If you want me to, I'm happy to integrate them myself. If you don't think they improve the logging infrastructure, I'm also fine with dropping that.

nbenn:
@DivadNojnarg could you check this out quickly. I expanded a bit on the logging infra as proposed by @JohnCoene. It defeats the purpose if I just self-approve so a quick review from you would be great.

@JohnCoene I hope what I did is ok with you. Let me know if you feel otherwise/want me to change things. I wanted to add you as reviewer, but apparently couldn't (b/c you opened the PR I presume).

A quick remark on implementation: Initially I had something like

```r
default_logger <- function(msg, level) {

  if (level == fatal_log_level) {
    stop(msg, call. = FALSE)
  }

  if (level == error_log_level) {
    out <- stderr()
  } else {
    out <- stdout()
  }

  cat(msg, file = out)
}
```

to mimic the behavior that you initially proposed. I then changed this (essentially to all stderr, using R conditions) b/c I believe on hosted shiny environments such as Connect or Shiny Server only output to stderr is picked up. Not entirely sure here, happy to further investigate, but changing this at a later point is easy (and it is even user-configurable) with what we have here, so one could for example set the above as `BLOCKR_LOGGER` option in their runtime env.

JohnCoene:
We can pipe to stdout, it's supported by Connect.

Originally, I opened this so we could silence the large number of messages when not running with `BLOCKR_DEV`, it seems I cannot silence the messages, or is that intentional?

nbenn:
@JohnCoene Sure thing.

I changed the default logger to follow what you proposed initially (i.e. write to stdout for warn, info, debug, trace levels and to stderr for non-fatal errors). This is user configurable and I kept the condition logger around, so this can be enabled in a setting where stdout causes issues. I'm pretty sure for Shiny Server this (at some point at least) was an issue, but it might no longer be these days. Even better if it never was for Connect.

As for what is shown and what is not: having `BLOCKR_DEV` as `TRUE` will set logging to "debug" and "info" otherwise. This is why all `log_info()` msgs are shown in that case. I swapped all of them out for `log_debug()`. Some might even be candidates for trace level.

DivadNojnarg:
@nbenn: I quickly tried the log system. It seems to make the app very laggy. When I am on trace/debug/info level, it takes 2-3 sec to render a 150 rows iris dataset, then adding a select block and changing a column takes about 5sec (so I don't know what will happen for more expansive operations). That's my main observation here.

nbenn:
@DivadNojnarg It might be due to the memory reporting I added last minute. I'll check it out tomorrow and perhaps disable it again. Thanks for the feedback!

nbenn:
@DivadNojnarg For me, there is a bit of a difference between including memory info in logs and not doing so. But it is not as drastic as you describe above. At any rate, I made it opt-in (via `options(BLOCKR_LOG_MEM = TRUE)`). Can you check if this improves things for you (i.e. if this explains the slow-down you observed)?

JohnCoene:
`BLOCKR_DEV` was intended to be a boolean so I assume this PR will break existing use of the option, e.g.:

https://github.com/blockr-org/blockr/blob/0c6728fcfe53bc9b4851a3c86e9799500ecf5c5d/R/ui.R#L176

We'd need to change this

nbenn:
@JohnCoene How does this break things? It surely is not the intention here to no longer have the `BLOCKR_DEV` option to be binary. If something is broken here, tell me what and I'll fix.

JohnCoene:
Ah sorry my bad, it's another option `BLOCKR_LOG_LEVEL`

DivadNojnarg:
@nbenn. Your last change works well on my side.

DivadNojnarg (approved):
Thanks for the changes. I tested the latest changes which solve the slowness issue. You can merge.

Issue #254, opened 2024-02-02
UI widget for last field duplicated


karmatarap:
The generated ui duplicates the creation of the last field. The value is non-reactive for this field

```r
library(shiny)
library(blockr.data)
library(blockr)

stack <- new_stack(data_block)

#' ggscatterstats
#'
#' A new transform block.
#'
#' @export
ggscatterstats_block <- function(data, ...) {
  types <- c(
    "parametric",
    "nonparametric",
    "robust",
    "bayes"
  )
  num_cols <- colnames(dplyr::select_if(data, is.numeric))
  blockr::new_block(
    expr = quote({
      ggstatsplot::ggscatterstats(data = data,
                                  x = .(x),
                                  y = .(y),
                                  type = .(type),
                                  #  conf.level = .(conf.level),
                                  xlab = .(xlab),
                                  ylab = .(ylab),
                                  title = .(title))
    }),
    fields = list(
      x = blockr::new_select_field(num_cols[1], num_cols),
      y = blockr::new_select_field(num_cols[1], num_cols),
      type = blockr::new_select_field(types[1], types),
      #  conf.level = blockr::new_numeric_field(0.95),
      xlab = new_string_field(""),

      title = new_string_field(""),      ylab = new_string_field("")
    ),
    class = c("ggscatterstats_block", "plot_block")
  )
}


blockr::register_block(
  constructor = ggscatterstats_block,
  name = "scatter plot",
  description = "scatter with stats",
  classes = c("ggscatterstats_block", "plot_block"),
  input = "data.frame",
  output = "data.frame"
)

shinyApp(
  ui = bslib::page_fluid(
    add_block_ui(),
    generate_ui(stack, id = "mystack")
  ),
  server = function(input, output, session) {
    vals <- reactiveValues(new_block = NULL)
    stack <- generate_server(
      stack,
      id = "mystack",
      new_block = reactive(vals$new_block)
    )

    observeEvent(input$add, {
      vals$new_block <- NULL
      # Always append to stack
      loc <- length(stack$blocks)
      block <- available_blocks()[[input$selected_block]]
      # add_block expect the current stack, the block to add and its position
      # (NULL is fine for the position, in that case the block will
      # go at the end)
      vals$new_block <- list(
        block = block,
        position = loc
      )
    })
  }
)
```

![image](https://github.com/blockr-org/blockr/assets/17053126/102e89d1-5afa-4f6d-b765-819cf6090d6b)

[...]

PR #256, opened 2024-02-02
fix: n fields check on layout

JohnCoene:
Quick fix for @karmatarap's issue in #254 

Simple sign issue: `<` changed to `<=`

### DIFF
--- a/test.R
+++ b/test.R
@@ -2,42 +2,81 @@ devtools::load_all()
 library(blockr.data)
 library(shiny)
 
-new_test_block <- function(data, ...) {
-  fields <- letters |>
-    lapply(\(x) new_string_field(x))
+stack <- new_stack(data_block)
 
-  names(fields) <- letters
-
-  new_block(
-    fields = fields,
+#' ggscatterstats
+#'
+#' A new transform block.
+#'
+#' @export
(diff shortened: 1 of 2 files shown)


test.R
```diff
@@ -2,42 +2,81 @@ devtools::load_all()
...
+#'
+#' @export
+ggscatterstats_block <- function(data, ...) {
+  types <- c(
+    "parametric",
+    "nonparametric",
+    "robust",
+    "bayes"
+  )
+  num_cols <- colnames(dplyr::select_if(data, is.numeric))
```
nbenn, line 19: If you want this to update according to data changes, you'd have to pass a function instead of a static character vector.

Something like

```r
num_cols <- function(data) {
  colnames(dplyr::select_if(data, is.numeric))
}
```

and then for example use as

```r
new_select_field(num_cols(data)[1L], num_cols)
```

nbenn (approved):
Lgtm.

test.R
JohnCoene: Thanks, this is an example I took from @karmatarap in https://github.com/blockr-org/blockr/issues/254

nbenn: Yeah, I saw and made a comment for @karmatarap there. Thanks.

PR #260, opened 2024-02-02
Allow downstream blocks after plot blocks

### LINKED ISSUE #259: Please allow blocks after plot blocks
I would like to use a function on ggstatsplot to extract stats from a statsplot object, can we please remove this line from stack.R
`     if (inherits(last, "plot_block")) {
       stop("Can't insert a block below a plot block.")
     }`

DivadNojnarg:
Fix #259.
Don't know what @JohnCoene and @nbenn you think but I see that adding more metrics after a plot block would make sense, if people want to show value boxes for instance.


DivadNojnarg:
TO DO: ensure that the plot returns the plot data and not the plot element.

R/stack.R
```diff
@@ -113,17 +113,6 @@ add_block <- function(stack, block, position = NULL) {
     }
   }
 
-  if (length(stack) > 0L) {
```
nbenn, line 116: @DivadNojnarg Removing such hard-coded logic is a good change I feel. The way forward in my opinion should be use of info available from the block registry.

https://github.com/blockr-org/blockr/blob/0f8d554b330d85be9148783cd869a476a56720cb/R/registry.R#L167-L168

That can be used to determine whether a block can be added in any given position (by checking if `input` of i matches output of i - 1.

This also goes for figuring out which blocks are allowed to "start" a stack (i.e. have `is.na(input)`).

DivadNojnarg: What happens if the registry metadata are not consistent, for instance we have `data.frame` instead of `NA_character_` as input for a data block. At the moment, we don't have control over what people put in the registry so relying on this would not be enough.

nbenn (approved):
Lgtm.

PR #264, opened 2024-02-04
Stack result re-use

### LINKED ISSUE #263: Use stack result in block
Stack results should be available as stack/block inputs:

- Add a field class that allows selecting results of all stacks in the workspace,
- use this field in a data block,
- and in a join block (#225)

nbenn:
Fixes #263.


nbenn:
@DivadNojnarg we have something here. It's not a super elegant solution but it is minimally invasive and more importantly, it works.

nbenn:
@DivadNojnarg good for review.

DivadNojnarg:
This is fun to see it working! I am going to change the join block, as discussed since it has to be fixed.

DivadNojnarg (approved):
LGTM. Thanks. 

A minor comment would be regarding the block result field. I initially removed the `y` parameter passed from outside. However, for testing, this is needed because the result field isn't populated when you do this in a testthat context:

```r
block <- join_block(
    band_members,
    type = "left"
  )
```
and the code will always be `identity()` because the result field isn't initialized. I had to re-include a `data_merge` parameter which allows to bypass this issue in a testing context. @nbenn If you have a more elegant solution, feel free to add it to this PR.

R/block.R
```diff
@@ -622,38 +622,37 @@ group_by_block <- function(data, ...) {
...
-      by_col
-    } else {
-      choices[[1]]
-    }
+  default <- function(data, y) {
+    if (length(by_col) == 0) by_choices(data, y)[1] else by_col
   }
 
+  # Capture user input
+  if (!is.null(data_merge)) data_merge <- substitute(data_merge)
```
nbenn, line 725: I don't think this is needed. The problem might be (not entirely sure, didn't check this) that you are running into an issue where `data_merge` is evaluated in a clean environment (where dplyr is not a parent). Might be wrong here.

R/block.R
```diff
@@ -622,38 +622,37 @@ group_by_block <- function(data, ...) {
...
+
```
nbenn, line 726: This entire logic is not needed I thing (and if it was, that'd be a bug).

- the `length()` checks should be covered by `is_initialized()`
- the hidden field is not needed here

inst/examples/connect-stacks/app.R
```diff
@@ -0,0 +1,28 @@
...
+
+stacks <- list(
+  new_stack(lab_data_block, head_block),
+  new_stack(ae_data_block, join_block)
+)
+names(stacks) <- paste("stack", 1:2, sep = "_")
+
+do.call(set_workspace, args = c(stacks, list(title = "My workspace")))
+
+serve_workspace(clear = FALSE)
```
nbenn: You can also set up the workspace by passing the stacks as args.

nbenn:
@DivadNojnarg I tried adding some improvements (not functional, but more to make things a bit more idiomatic) to the join block. I checked your "connect-stacks" example and it seems ok to me.

Please have a quick look at my changes (I also added some comments above to explain a bit). If you have questions, I'm happy to discuss quickly.

Also maybe quickly check that everything still functions as you intended. I hope I didn't break anything.

DivadNojnarg:
@nbenn Thanks for those changes. Two comments:
- `by`'s default value is `NULL`. This can work, if we assume the user wants to select the `by` her/himself.
- We don't fully handle the case where the selected `y` isn't compatible with the input data. Currently, you can crash the app  in a dynamic context (I also don't think the `by` is correctly updated). Example below: my stack contains 2 incompatible data blocks upon instantiation and the `by` will be empty. This is what we expect:
```r
library(blockr)
library(blockr.data)

lab_data_block <- function(...) {
  initialize_block(new_data_block(
    ...,
    dat = as.environment("package:blockr.data"),
    selected = "lab"
  ))
}

serve_workspace(
  stack1 = new_stack(lab_data_block, head_block),
  stack2 = new_stack(data_block, join_block),
  title = "My workspace"
)
```

Now, if I start the app with 2 compatible datasets, I add a third stack with a `data_block`, not compatible with the second stack data set, select this new stack as my new `y` value, the `by` choice doesn't reflect the situation and keeps the old values in memory. The result is shown in the below image where I am trying to join adverse events data with the airquality data (which doesn't make sense of course).

```r
library(blockr)
library(blockr.data)

lab_data_block <- function(...) {
  initialize_block(new_data_block(
    ...,
    dat = as.environment("package:blockr.data"),
    selected = "lab"
  ))
}

ae_data_block <- function(...) {
  initialize_block(new_data_block(
    ...,
    dat = as.environment("package:blockr.data"),
    selected = "ae"
  ))
}

serve_workspace(
  stack1 = new_stack(lab_data_block, head_block),
  stack2 = new_stack(ae_data_block, join_block),
  title = "My workspace"
)
```
Screenshot 2024-02-07 at 12 08 35

PR #266, opened 2024-02-05
More complete workspace update utils

### LINKED ISSUE #221: Workspace methods and handling
This relates to #129 

1. I think we should add methods to the workspace (or functions to manipulate it). Currently we essentially only have one function to "set the workspace" it would be good if we could add some CRUD methods.

Generally the user adds, updates, or removes a single stack not multiple stacks at once so having these methods would help.

2. I think most of the workspace handling could be internalised within {blockr}, we need to update the workspace when a block is added, removed to a stack or when a block is updated. All of this can be observed within blockr but I currently have to handle that outside of blockr.

nbenn:
@JohnCoene A minor (backwards compatible) update to the workspace API. We now have

- `rm_workspace_stacks()`, in addition to the previous `rm_workspace_stack()` to remove a set of stacks
- `set_workspace_stack()` to update an existing stack (previously you could do `add_workspace_stack(., ., force = TRUE)` but it seems more explicit to be able to change an existing stack which warns when the stack does not exist, whereas adding with `force` would then just add

Let me know if you would like something additional here. If not, this closes #221.


DivadNojnarg (approved):
Thanks!

PR #274, opened 2024-02-07
Update by in join block

### LINKED ISSUE #273: Dataset switch in join block does not update `by`
If I start the app with 2 compatible datasets, I add a third stack with a `data_block`, not compatible with the second stack data set, select this new stack as my new `y` value, the `by` choice doesn't reflect the situation and keeps the old values in memory. The result is shown in the below image where I am trying to join adverse events data with the airquality data (which doesn't make sense of course).

```r
library(blockr)
library(blockr.data)

lab_data_block <- function(...) {
  initialize_block(new_data_block(
    ...,
    dat = as.environment("package:blockr.data"),
    selected = "lab"
  ))
}

ae_data_block <- function(...) {
  initialize_block(new_data_block(
    ...,
    dat = as.environment("package:blockr.data"),
    selected = "ae"
  ))
}

serve_workspace(
  stack1 = new_stack(lab_data_block, head_block),
  stack2 = new_stack(ae_data_block, join_block),
  title = "My workspace"
)
```
Screenshot 2024-02-07 at 12 08 35

_Originally posted by @DivadNojnarg in https://github.com/blockr-org/blockr/issues/264#issuecomment-1931816336_

nbenn:
Fixes #273, currently at the expense of breaking the filter block. Working on a solution to have both.

### DIFF
--- a/R/field.R
+++ b/R/field.R
@@ -92,10 +92,7 @@ initialize_field.field <- function(x, env = list()) {
 eval_set_field_value <- function(x, env) {
   for (cmp in names(x)[lgl_ply(x, is.function)]) {
     fun <- x[[cmp]]
-    tmp <- do.call(fun, env[methods::formalArgs(fun)])
-    if (length(tmp) > 0) {
-      value(x, cmp) <- tmp
-    }
+    value(x, cmp) <- do.call(fun, env[methods::formalArgs(fun)])
   }
 
   x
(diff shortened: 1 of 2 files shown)


DivadNojnarg:
There's also something about the filter expression: it is including `call(`+`, ..., ...)`. I wrote this long time ago so there might be an easy way to make it nicer, that is not like this:

```r
get("ae", envir = as.environment("package:blockr.data")) %>%
        dplyr::filter(eval(call("==", DOMAIN, "AE")))
```

PR #279, opened 2024-02-08
Resolve crash on dataset change

nbenn:
@DivadNojnarg should be fairly safe (I hope), essentially just adding a try catch that wraps user supplied functions used as field components.

### DIFF
--- a/R/field.R
+++ b/R/field.R
@@ -90,10 +90,19 @@ initialize_field.field <- function(x, env = list()) {
 }
 
 eval_set_field_value <- function(x, env) {
+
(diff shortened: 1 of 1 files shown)


DivadNojnarg (approved):
The app does not crash anymore indeed. Still a little issue with the slider, for which the old string field should be hidden because columns is empty (currently shown with the old value), but that's not the purpose of this PR.
Screenshot 2024-02-08 at 11 51 10

nbenn:
@DivadNojnarg yeah, I saw the missing update to the input title too. It does update as soon as you select a new column and I think for now this is the best we can do.

PR #280, opened 2024-02-08
fix: hide loading on plot blocks

JohnCoene:
This is a simple one line change to hide the loading icon on plot blocks too


JohnCoene:
@DivadNojnarg could we merge that? This is really a one line change at its core.

Changing this `const tableId = $lastOutput.find(".datatables").first().attr("id");` for `const tableId = $lastOutput.find(".shiny-bound-output").first().attr("id");`

DivadNojnarg (approved):

PR #282, opened 2024-02-08
Fix blank plot

JohnCoene:
Changes the logic of rendering blocks from stack. We can discuss next week but this works for now.

I would argue it's in part better. before we were rendering all blocks as collapsed/hidden only to render the last block on load.
Here we check the block index in the stack so we render the last block as visible.
This seems to fix the issue: shiny can behave strangely when inputs/outputs are hidden as it tries to ignore them.

It's not a super elegant fix as some part of the last block are being hidden via JavaScript on render: this should be changed.


nbenn (approved):
Works for me @JohnCoene, thanks!

Issue #245, opened 2024-01-31
Delete stack


JohnCoene:
In any application we'd need some functionality to delete any single stack.

I'm split about whether it should be within blockr or elsewhere, currently not being handled.

nbenn:
As with many such things, I feel that blockr should provide the basic infrastructure to handle this. In principle, several things are already available:

- `rm_workspace_stack()` to manipulate the workspace
- Functionality in the workspace server function such as 
    https://github.com/blockr-org/blockr/blob/4124a23cfbb3121a71b93299707f7b10f749fc0b/R/server.R#L486

Now we currently have some issues with having the stack/workspace level server functions properly update (tracked in #234), but apart from that, are there things that we are missing?

JohnCoene:
Agree that it should be in blockr.

UI:
- delete button

Server:
- removeUI
- remove from workspace
- reset inputs (optional)

---

I think we already have most required components: we just need a new button/link to delete in `.stack-tools`

https://github.com/blockr-org/blockr/blob/4124a23cfbb3121a71b93299707f7b10f749fc0b/R/ui.R#L355

And the corresponding server logic that can just `removeUI` and (optionally) remove/reset stack inputs (I rarely reset input values but @DivadNojnarg may have a different option, I think we do do that on block removal).

JohnCoene:
Ah yes indeed, I forgot that.

It's something I had put in place for the demo and would have to be either removed or changed.

It's not good as it is, it makes the assumption that it uses masonry, etc.

`removeUI` is probably easier.

@DivadNojnarg do we need to propagate this to the blocks. We reset the inputs for the blocks when we remove a block, if we remove a stack should we reset the inputs of the children blocks too?

DivadNojnarg:
No we don't reset because IDS are randomly generated and that does not seem to cause issues so far.

DivadNojnarg:
In that case we could clean the JS code (which is probably easier if we assume blockr developers will mostly have a R background).

JohnCoene:
I see some issues with how this is currently handled.

`inject_remove_button` is never called in `generate_ui.stack`

https://github.com/blockr-org/blockr/blob/0a4696821d9c48992131c69ab66bfbf6bd72dfd4/R/ui.R#L318

We build the UI differently from the workspace.

https://github.com/blockr-org/blockr/blob/0a4696821d9c48992131c69ab66bfbf6bd72dfd4/R/ui.R#L394

PR #298, opened 2024-02-13
245 delete stack

JohnCoene:
This PR was initiated to add the delete button to the stack #245 but I uncovered a few issues.

I don't think we have to "inject" the stack delete button, it should just be part of the UI.
The removal of stacks is handled at the workspace level, which seems strange to me; the stack "injects" the button in itself but the server code sits on the workspace server.
@DivadNojnarg may I change that? Is there something I am missing?

This PR starts this but is likely breaking things on the workspace server.


R/server.R
```diff
@@ -359,6 +305,24 @@ generate_server.stack <- function(x, id = NULL, new_block = NULL,
...
 
+      observeEvent(input$remove, {
+        removeUI(
+          sprintf("#%s", session$ns(NULL))
+        )
+
+        rm_workspace_stack(id, workspace = workspace)
+      })
+
+      observeEvent(input$add, {
```
DivadNojnarg, line 316: What's the difference between 
```r
observeEvent(
        {
          req(new_block)
          new_block()
        },
        {
          add_block_stack(new_block()$block, new_block()$position, vals = vals)
        }
      )
```

and the `input$add` observer? Is it to handle the case where you have drag and drop?

R/server.R
```diff
@@ -700,3 +663,63 @@ server_code.block <- function(x, state, output) {
     cat(deparse(generate_code(state())), sep = "\n")
   )
 }
+
+
+add_block_stack <- function(
```
DivadNojnarg, line 668: Thanks for reducing the observer code.

R/ui.R
```diff
@@ -395,6 +388,14 @@ stack_header.stack <- function(x, title, ns, ...) {
...
         div(
           class = "stack-tools",
+          actionLink(
+            ns("remove"),
+            class = "text-decoration-none stack-remove",
+            `data-bs-toggle` = "tooltip",
+            `data-bs-title` = "Remove stack",
+            icon("trash")
+          ),
+          add_block_ui(ns),
```
DivadNojnarg, line 398: Do you remove this piece of logic in the `blockr.demo`, as you're bringing blocks by drag and drop action?

DivadNojnarg (commented):
I tested, the stack level function which seems to work fine. While quickly testing the workspace app, I end up with few issues: 
- Removing a stack does not update the number of stacks captured in `length(vals$stacks)`. 
- There're few floating tooltips remaining on the ui.
Screenshot 2024-02-16 at 15 36 46

R/ui.R
JohnCoene: I can always programmatically hide it from outside blockr if wanted. I don't see a problem with having both. 

I would like to improve this add block UI at a later stage after this is merged.

DivadNojnarg: I am fine if it's not a problem for you. There are rooms for improvement regarding this piece of UI, it's pretty ugly ;). Feel free to improve in another PR.

JohnCoene:
@nbenn last two commits fix the two remaining issues, good to review.

JohnCoene:
Note on bbd4c183ac18404732abe1ef61be1e9c9a3c5482 I'm not sure it's best like that or to remove entirely.

JohnCoene:
@DivadNojnarg @nbenn can we look into merging this?

DivadNojnarg:
Will check later today

DivadNojnarg (commented):
With the latest changes, tooltips correctly disappear. Minor comment: The layout is still quite strange. Feel free to fix if you can.
To reproduce:

```r
# library(blockr)
devtools::load_all()

do.call(set_workspace, args = c(list(), list(title = "My workspace")))

```

- Clear all stacks (button clear all)
- Add 3 empty stacks
- Remove the second stack: the third stacks remains in the third position, which leaves  a big gap.

JohnCoene:
I pushed something to fix it.

DivadNojnarg (commented):
There seem to be a remaining issue:

- Adding stacks 1 by 1: the first 2 stacks are vertically added and the other one are added horizontally to the right of the first stack (See video)

https://github.com/blockr-org/blockr/assets/18291543/dfb33d13-cf78-41fc-88ad-de693fa77f4d

JohnCoene:
Alright, sorry about that. I fixed the issue and switched to using flex instead of Bootstrap's grid.

R/server.R
```diff
@@ -442,24 +442,13 @@ generate_server.workspace <- function(x, id, ...) {
...
 
         stack_ui <- div(
-          class = "col stack-col",
+          class = "flex-grow-1 stack-col m-1",
           generate_ui(el, session$ns(stack_id))
         )
 
         insertUI(
-          selector = if (length(vals$stacks) == 0) {
-            ".workspace"
```
DivadNojnarg, line 497: @JohnCoene: removing this seems to break when we clear all stack and try to add a new stack. Nothing appear in the UI.

JohnCoene: Indeed, fixed in the clear stacks observer instead https://github.com/blockr-org/blockr/pull/298/commits/409918f2a9f8b5876513b71efb55904af6416ff1

DivadNojnarg (approved):
LGTM. Thanks.

PR #318, opened 2024-02-17
Collapse block on add fixes #310

### LINKED ISSUE #310: Collapse block on add
One thing that appears annoying as I observe people use blockr is having to constantly scroll as they build stacks.

I suggest we collapse the previous block when we add one so it keeps the stack more compact as it is being built. There's also probably no need to show the output of a block once we've move to working on the next, the previous block is not longer the user's concern anyway.

@karmatarap what do you think?

JohnCoene:
This collapses all blocks when one is added.


JohnCoene:
This now does indeed collapse all blocks except the added block, makes it easier to use as users progressively add blocks to build their pipeline/visualisation.

It also fixes some inconsistencies in the collapsed state/icons of blocks (which was necessary to get the "collapse on add" working properly)

Also @DivadNojnarg it refactors a bit the add block see `server.R` in https://github.com/blockr-org/blockr/pull/318/commits/6fc538a3fa66141b153f29b43f6f3dcb2b03ad06 this is so we have a unified interface/unique function to add block to stack.

DivadNojnarg (approved):
Thanks @JohnCoene LGTM. The new change works well. 

Notes for a later PR:
- I still see an inconsistent state of the blocks collapse icons, after toggling the stack multiple times. As this problem was here before, I don't want to delay this PR.
- Add and remove block buttons shouldn't be next to each others, at least. I'd even argue that the remove button should be subject to a confirmation modal (as any destructive action, UI/UX best practices. We can check this @karmatarap). This can be done in a later PR.

JohnCoene:
OK, thanks, I'll look at the other issues.

PR #325, opened 2024-02-28
Add blocks

JohnCoene:
This PR adds an "add block" button to stacks, it ports over features from {blockr.ui}.

The button to add blocks is hidden when the stack is collapsed.
The add block functionality reuses some of the previous implementation but is server-side rendered (infinite scroll + search) to support the growing number of blocks.

EDIT: this branched off #298


R/utils.R
```diff
@@ -413,3 +415,34 @@ create_app_link <- function(app_code, mode = c("app", "editor"), header = TRUE)
...
+  return(4)
+}
+
+sort_registry <- function(blocks) {
+  blocks[order(sapply(blocks, registry_sorter), na.last = FALSE)]
+}
+
+add_block_index <- function(blocks) {
+  for (i in seq_along(blocks)) {
+    attr(blocks[[i]], "index") <- i
```
DivadNojnarg, line 440: I think block already have a position attribute. Can't this be used here? (Or am I missing something).

DivadNojnarg (commented):
Screenshot 2024-02-28 at 23 31 22

Thanks for reworking this. Few comments sorted by decreasing importance:

- Now add and delete are side by side. It is easy to accidentally delete the stack while trying to add a block.
- After searching for a block and reseting the search field, search should be reset. I currently have to click again on search to reset it.
- Is there any possibility to sort pills by type (data, wrangling, ... plot): I acknowledge the number of blocks is rather limited in the core and we don't have plot blocks.

JohnCoene:
To search type then enter, we have to execute `fetch` calls to search so I think it's wiser like this. To reset set search clear and hit enter.
Blocks are already sorted by type

DivadNojnarg:
- json, sas, rds, ... are data parsers so they should belong to data blocks (?). What I meant by sorting blocks, sorry for not being more accurate in my comment: in addition to the color scheme we can perhaps add a subtitle for each category (thinking about someone new to R and dplyr).

- I also realised that we probably don't want to `result` block (labelled as stack) in a single stack because it is assumed to work with the workspace and multiple stacks.

- > To reset set search clear and hit enter

As a user, I would expect the search to auto reset when I clear it (without having to click on enter).

R/utils.R
JohnCoene: That's the index in the registry, not in the position in a stack-

JohnCoene:
The search now runs on keyup, it's likely to affect performance though.

I sort the blocks but categorising them is more difficult given they may fall in a category outside of "data", "transform", "visualise" (this depends on classes and users are free to extend this)

JohnCoene:
This branched off #298 so I would not advise merging this before #248

R/registry.R
```diff
@@ -279,3 +279,193 @@ construct_block <- function(block, ...) {
...
+      )
+    )
+  )
+}
+
+add_block_server <- function(
+  session,
+  registry = available_blocks
+) {
+  if (!getOption("BLOCKR_ADD_BLOCK", TRUE))
```
nbenn, line 356: Why do we need this env var? If the stack-level "add-block" component is disabled, how can blocks be added (using the default UI)? If this is something you want to be able to *override* in a more sophisticated UI implementation, we should think about infrastructure for having some "hooks" in certain places to enhance default behavior.

nbenn:
@JohnCoene a minor remark regarding the current UI: why is there so much whitespace between the block pills and the block "info box"? How about providing this info in some pop-over?

R/registry.R
```diff
@@ -279,3 +279,193 @@ construct_block <- function(block, ...) {
...
+    unname()
+
+  shiny::httpResponse(
+    200L,
+    content_type = "application/json",
+    content = jsonlite::toJSON(blocks, auto_unbox = TRUE, dataframe = "rows", force = TRUE)
+  )
+}
+
+search_registry <- function(data, req) {
```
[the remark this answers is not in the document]
JohnCoene: Here we only fetch the data available when requested. A dashboard with many stacks would need to send the entirety of the registry for every stack which does not scale.

A fetch call for a search to server-side render is pretty standard, e.g.: sielectize.js

R/registry.R
JohnCoene: I just thought the ability to turn that off for users that only want to use a custom UI would be nice.

What do you mean by "hooks"?

R/registry.R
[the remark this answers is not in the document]
nbenn: The number of objects here is the number of available block "types", so this number will not become massive anytime soon, don't you agree? Even if at some time, there exists several hundred blocks, the entirety of the serialized registry information for the frontend is negligible. If we're talking about >1e4-1e5 blocks, then yeah, subsetting that might be smart. Don't you agree?

With the current implementation you first retrieve all block. Then for each search, you round trip to fetch a subset of what you already loaded initially, making every subsequent round-trip pointless. Or am I somehow misunderstanding this?

Also, the set of known blocks is not "stack-level" information. This is fixed for all stacks, so if you fetched all initially, you could re-use this for all stacks instead of fetching per stack, no?

R/registry.R
nbenn: What I meant was that if we want to have certain components be overridable, we should have some basic API for this, that is consistently used. Something like (for this example) having default `add_block_ui()` and `add_block_server()` and then making it possible to override this by an approach like

- passing arguments in `generate_ui()` and `generate_server()` (probably not an approach that scales well),
- have some config object (like the registry) where you can set these overrides,
- more in line with what we're doing elsewhere, we could have an S3 class for the UI and then dispatch functions like `stack_header()` or the add-block stuff on that.

Maybe you have other ideas? I'm just trying to formalize the underlying problem a bit, as this pops up in many places.

JohnCoene:
Refactored everything to reduce number of calls to server, ready for review.

R/utils.R
```diff
@@ -413,3 +415,37 @@ create_app_link <- function(app_code, mode = c("app", "editor"), header = TRUE)
     `data-external` = "1"
   )
 }
+
+registry_sorter <- function(block) {
```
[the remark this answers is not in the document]
JohnCoene: The idea of the sort blocks in the order we use them

1. data
2. parser
3. transform
4. visualise

For the categories to appear in the order the are in the registry does not really make sense to me, e.g.: currently it would start with parser blocks.

We kind of have to hard-code that for now at least, otherwise how do we define a "category"? I know it's not ideal but it is fair to say there is an implied category that allows sorting the blocks to present them in a way that is more navigable and understandable to the user.

R/utils.R
```diff
@@ -413,3 +415,37 @@ create_app_link <- function(app_code, mode = c("app", "editor"), header = TRUE)
...
+  if ("transform_block" %in% cl)
+    return(3)
+
+  if ("plot_block" %in% cl)
+    return(4)
+
+  return(5)
+}
+
+sort_registry <- function(blocks) {
```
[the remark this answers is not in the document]
JohnCoene: The sorting I have already does point 2 normally, the sorting here only changes order of "categories."

See my response to your previous comment, I think the point of sorting here is precisely to have it in a specific order. Besides, the implicit order would then depend on the order in which the user calls `library` of blockr packages:

```r
# block in an order
library(blockr.extra)
library(blockr.ggplot2)

# block in a different order
library(blockr.ggplot2)
library(blockr.extra)
```

JohnCoene:
I've pushed a few changes to simplify and the logic and have use a reactive registry.

This is because we are going to need to be able to push to the registry from within the app:

- A module to create blocks from the app (low priority)
- An llm block that itself creates blocks

```r
library(blockr)
library(blockr.data)
library(blockr.ggplot2)

stack <- new_stack()

shiny::shinyApp(
  ui = bslib::page_fluid(
    actionButton("load", "load package"),
    generate_ui(stack)
  ),
    generate_server(stack)
    observeEvent(input$load, {
      library(blockr.ggstatsplot)
    })
  }
)
```

Issue #328, opened 2024-03-05
Validation - server-side


JohnCoene:
Some observations/questions on server-side validation of inputs, maybe not so much the validation itself but how it is handled within the shiny modules.

Server side we create the `is_valid` reactive but never react on any of its changes.

https://github.com/blockr-org/blockr/blob/78782f38edd9c45be9ee012dc4c90149e784853b/R/server.R#L64

The object above is not used for plot blocks.

https://github.com/blockr-org/blockr/blob/78782f38edd9c45be9ee012dc4c90149e784853b/R/server.R#L127

The `validate_inputs` function accepts the `is_valid` reactive list but resets it anyways

https://github.com/blockr-org/blockr/blob/78782f38edd9c45be9ee012dc4c90149e784853b/R/utils.R#L293

Also note that `is_valid` has the `inputs` entry and `validate_inputs` and resets the `inputs` entry of the list but the `validate_inputs` functionally internally loops over inputs to valid and assigns it to `input` (not `inputs`) 

-----

I'm exploring this as I look at how we can make things more robust and avoid errors in blockr crashing applications, (#303 #304)

DivadNojnarg:
All of this has been broken when the block module got refactored. I had to disable it in plots as it was preventing to add the layers. If you work on this, I'd suggest to restart from scratch, as other pieces are not used anymore like `error`. 

The bigger questions is how to make this scalable to stacks and workspace, when stacks are connected.

DivadNojnarg:
Fixed in 0.0.2

PR #329, opened 2024-03-05
Error handling

JohnCoene:
This PR started as a look into preventing app crashes on error, hence the branch name but then went into looking error handling.

Some of this was discussed with @DivadNojnarg in #328 .

This is what I would propose, create a stack of error messages which we eventually handle.

There would still be lots to do but I leave it at this for feedback before needlessly delving into it if we think this should be done some other way.


JohnCoene:
Note to self, there is a `secure` function in utils (wrapper around `tryCatch`)

JohnCoene:
Will redo as part of validation refactor

Issue #332, opened 2024-03-06
Fields description


JohnCoene:
Need to handle fields description, currently not being used. Tooltips?

ppssphysics:
Was implemented in #413.

PR #333, opened 2024-03-06
Field titles

### LINKED ISSUE #314: Filter block field order
Change order of fields in filter block so they display in a more logical manner.

column -> selector -> value

### LINKED ISSUE #319: Field label
Add ability to set field display label

JohnCoene:
Use field `title` attribute as label for inputs (if set).

This closes #319 and fixes #314 by the way

Not dealing with #332 yet.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -206,7 +206,11 @@ new_data_block <- function(
   if (length(selected) == 0) selected <- datasets[1]
 
   fields <- list(
-    dataset = new_select_field(selected, datasets)
+    dataset = new_select_field(
+      selected,
+      datasets,
+      title = "Dataset"
+    )
   )
 
   expr <- substitute(
@@ -238,7 +242,7 @@ new_upload_block <- function(...) {
 
   new_block(
     fields = list(
-      file = new_upload_field(),
+      file = new_upload_field(title = "File"),
       expression = new_hidden_field(data_path)
     ),
     expr = quote(c(.(expression))),
@@ -271,7 +275,7 @@ new_filesbrowser_block <- function(volumes = c(home = path.expand("~")), ...) {
 
   new_block(
     fields = list(
-      file = new_filesbrowser_field(volumes = volumes),
+      file = new_filesbrowser_field(volumes = volumes, title = "File"),
       expression = new_hidden_field(data_path)
     ),
     expr = quote(c(.(expression))),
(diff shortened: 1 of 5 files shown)


R/block.R
```diff
@@ -206,7 +206,11 @@ new_data_block <- function(
   if (length(selected) == 0) selected <- datasets[1]
 
   fields <- list(
-    dataset = new_select_field(selected, datasets)
+    dataset = new_select_field(
+      selected,
+      datasets,
+      title = "Dataset"
```
DivadNojnarg, line 212: I would put `title` as a parameter in case people want to change the default value.

R/field.R
```diff
@@ -557,3 +557,20 @@ generate_server.result_field <- function(x, ...) {
     })
   }
 }
+
+get_field_name <- function(field, name = "") {
```
DivadNojnarg, line 561: Add unit test to `get_field_name` and `get_field_names`.

JohnCoene: added tests

R/block.R
JohnCoene: added

R/block.R
```diff
@@ -714,7 +735,13 @@ new_group_by_block <- function(data, columns = colnames(data)[1], ...) {
 
 #' @rdname new_block
 #' @export
-group_by_block <- function(data, ...) {
+group_by_block <- function(
+  data,
+  ...,
+  field_join_func = "Type",
```
DivadNojnarg: With multiple fields, this does not look super nice, too many parameters. I don't have a good working idea on how to parametrise fields (at least field names) from the block. If we use a single `field_names` param which can take a vector of name, this might be misleading for single field blocks.

DivadNojnarg: Do you have better idea @nbenn ?

nbenn: Honestly (and sorry @JohnCoene for suggesting to revert a change you were asked to make), my proposal here would be to not expose these names as arguments for now. If we want to make such aspects customizable, I think we need to do something else.

@christophsax at one time floated the idea to also have other field-level metadata, such as a description, and I don't think it makes sense to surface all of that as block constructor arguments.

nbenn (commented):
@JohnCoene thanks for the tests. I don't have much to add apart from maybe removing the field titles as block constructor arguments and just hard-code them for now. No strong opinions though.

An advantage of not exposing this is that we could (not now, but medium term) wrap these things in `gettext()` calls to make them translatable. Apart from that I don't see much need to customize this (give that we have sensible names in place).

JohnCoene:
@nbenn no worries, I agree with you.

@DivadNojnarg I reverted that.

DivadNojnarg (approved):
Thanks @JohnCoene. Sorry for the wrong advise. Good to merge.

Issue #315, opened 2024-02-16
Lock - Expose inputs


JohnCoene:
Should we expose some inputs when the stack is "locked." Perhaps the dashboard builder still wants to allow the user to play with some inputs and explore the data? Ideally the dashboard builder would be able to hand pick the inputs that are editable when locked.

Sounds like a consequential change but it may be something we want to explore

JohnCoene:
Proposal is in branch `315-expose-inputs` (really basic, just a proposal).

We add an argument to `new_field` to allow controlling whether the input is shown or not on lock.

Then when the user locks the dashboard we look for these inputs and move them at the top of the stack.

![image](https://github.com/blockr-org/blockr/assets/7833796/e4698197-d69c-4617-8fd0-6812f6f10d7b)

I think the solution works well enough personally but the real problem is one we recently encountered in my PR to use field titles: how do we expose block/field arguments to the users?

In this proposal whether the input is shown or not when locked is hardcoded/defined by the block developer which is sub-par.

PR #337, opened 2024-03-11
Expose Inputs

JohnCoene:
This PR addresses #315: exposing inputs when the dashboard is locked.

image

This is a custom Shiny input, by default the input is hidden when blockr is locked.

If the input is selected to be shown when locked it will be displayed above blocks in the stack on lock. This is stored in a `always_shown` attribute on the field.

image

See `inst/examples/lock/app.R` for an example.

```r
library(blockr)

stack <- new_stack(data_block)
id <- "mystack"

shiny::shinyApp(
  ui = bslib::page_fluid(
    shiny::actionButton("lock", "Lock stacks"),
    generate_ui(stack, id)
  ),
    shiny::observeEvent(input$lock, {
      lock()
    })

    vals <- reactiveValues(new_block = NULL)
    stack <- generate_server(
      stack,
      id,
      new_block = reactive(vals$new_block)
    )

    observeEvent(input$add, {
      vals$new_block <- NULL
      # Always append to stack
      loc <- length(stack$blocks)
      block <- available_blocks()[[input$selected_block]]
      # go at the end)
      vals$new_block <- list(
        block = block,
        position = loc
      )
    })
  }
)
```


JohnCoene:
I would prefer to place the input within the label but it seems not to work, see issue and PR I opened.

https://github.com/rstudio/shiny/issues/3995
https://github.com/rstudio/shiny/pull/3996

JohnCoene:
This now seems to work just fine for me, feel free to let me know what you think of this proposed solution.

DivadNojnarg:
Thanks for looking into this.

If I understood correctly, the blockr app admin would manually toggle inputs users want to see, lock the stack and share the link with users? 

I am concerned about the way to scale to a real use case, say a workspace with 5 stacks each having 5 blocks and 2-3 inputs per block. Do we want to have/expose a programmatic way/api to initialise the workspace with tagged/untagged inputs so we don't have to manually toggle them? 

We have a `settings` param available for the workspace, which could be used to pass such configuration down to the stack/block/field. 

The difficulty I see right now is how to efficiently route which setting to which stack/block/field, as block ids are randomly generated. In a near future, as we plan to be able to create block outside the stack with PR #367, we can also control which name we give to them when we instantiate (from branch `110-1-no-data`): 

```r
blk <- dataset_block(name = "plop")
blk
$dataset
$value
[1] "airquality"

$choices
 [1] "airquality"       "anscombe"         "attenu"           "attitude"
...          

$multiple
[1] FALSE

attr(,"type")
[1] "literal"
attr(,"class")
[1] "select_field" "field"       
attr(,"title")
[1] "Dataset"
attr(,"descr")
[1] ""
attr(,"status")
[1] "active"
attr(,"exclude")
[1] FALSE

attr(,"name")
[1] "plop" <BLOCK_ID>
attr(,"expr")
get(.(dataset), envir = as.environment("package:datasets"))
attr(,"class")
[1] "dataset_block" "data_block"    "block" 
```

Assuming, we can wait for PR #367 and another PR to support instantiating a stack from a list of blocks (in addition to block constructors), we could then do something like this:

```r
blk <- dataset_block(name = "plop")

stacks <- lapply(
    1:3, 
    \(x) new_stack(blk, name = sprintf("stack%s", x))
)

do.call(
    set_workspace, 
    args = c(
        stacks, 
        list(
            title = "My workspace",
            settings = list(
                inputs_state = list(
                    "stack1-plop-dataset" = "locked",
                    "stack2-plop-dataset" = "locked",
                    "stack3-plop-dataset" = "unlocked" # OR leave NULL
                )
            )
        )
    )
)

```

Then you'd need some logic at the block/field level (?) to handle the `inputs_state` list and see whether the current field needs to be tagged.

Do you have suggestions @nbenn?

Issue #19, opened 2023-10-04
Testing: add shinytest2


DivadNojnarg:
`testServer` is not enough to test the app in its entirety...

DivadNojnarg:
At the same time, this won't increase the coverage: https://github.com/rstudio/shinytest2/issues/268

PR #346, opened 2024-03-21
Set timeout

### LINKED ISSUE #342: Spinner
Ensure spinner is present, should already be there but it was requested by a user (did not work?)

https://github.com/blockr-org/blockr/blob/9c8a75f0132dfb3abb78a07ab204d12211cfb815/srcjs/collapse.js#L114

JohnCoene:
This refactors a number of things on the front-end and fixes a few issues.

We've had many `setTimeout` on the front-end that cause issues when trying to restore a workspace: we can never be sure when things are rendered. It's just bad practice (my fault, I put them there). This PR removes them.

This PR also adds a check on the delete stack via a modal to request confirmation from the user.

Finally, it fixes #342 so we have loading spinners whilst blocks render (feature request).


JohnCoene:
(lack of coverage is raised by the addition of the modal for the stack removal, I'm not sure we can test without shinytest2)

srcjs/stack-title.js
```diff
@@ -1,29 +1,45 @@
...
-};
 
-const handleStackTitle = (stack, ns) => {
-  $(stack).find(".stack-title-input").off("keydown");
+  $("body").on("click", ".stack-title-save", (event) => {
+    const $group = $(event.target).closest(".stack-title");
+
+    const v = $group.find(".stack-title-input").val();
+    if (v === "") {
+      window.Shiny.notifications.show({
```
DivadNojnarg, line 15: @JohnCoene: I alway see the notification when I change the title, empty or not empty

Screenshot 2024-03-21 at 16 37 01

R/server.R
```diff
@@ -352,6 +352,42 @@ generate_server.stack <- function(x, id = NULL, new_block = NULL,
...
+            title = "Remove stack",
+            p("Are you sure you want to remove this stack?"),
+            div(
+              class = "d-flex",
+              div(
+                class = "flex-grow-1",
+                actionButton(
+                  session$ns("cancelRemove"),
+                  "Cancel",
+                  class = "bg-success",
```
DivadNojnarg, line 366: I would put the cancel in default color (grey/light). It can be confusing in green.
Screenshot 2024-03-21 at 16 39 33

DivadNojnarg:
@JohnCoene: Yes you can test the modal with shinytest2 but this won't be counted in the coverage %: https://github.com/rstudio/shinytest2/issues/268. That said, I am not saying we shouldn't test it :). I'll have to redesign the testing strategy in another PR so no worries for this PR.

JohnCoene:
Yes, I mentioned that because we have https://github.com/blockr-org/blockr/issues/19 :)

srcjs/stack-title.js
JohnCoene: This PR does not change that functionality I believe, this is changed in the 327-block-title PR where I can't reproduce that error. let me know if I am missing something

DivadNojnarg: Looking at the code, this notification isn't in the main branch and the "block title PR" is solving something different. I don't get it.

DivadNojnarg: The issue only happens when I click the green send button. If I press enter, the notification is behaving correctly.

R/server.R
JohnCoene: I've changed it in the interest of moving forward with this but I think it's better green: it's the standard approach for modal/steps where we stop the user to force a decision and green == safe and red == danger.

srcjs/stack-title.js
JohnCoene: Thanks for spotting that! Just pushed a fix.

DivadNojnarg (approved):
LGTM Thanks.

Issue #78, opened 2023-11-17
Explicit block parameters when calling new_stack


DivadNojnarg:
Right now, when we call `new_stack`, its not clear what are the default block parameters or we might want to change them:

```r
new_stack(
  data_block,
  filter_block
)
```

We currently have to go back to the code base or use autocompletion to get an idea of what the block function is doing.

New:

```r
new_stack(
  data_block(data = mtcars, ...)
  filter_block(column = "mpg", value = 21.0 , ...),
 ...
)
```

JohnCoene:
I think this suggestion is good and probably necessary.
E.g.: we will also need a way to restore a stack to specific settings in the future; this is an elegant way of going so.

DivadNojnarg:
Note for future me:

using `as.list(substitute(...()))`  allows to get a list of calls such as:

```r
[[1]]
data_block(dat = as.environment("package:datasets"), selected = "iris")

[[2]]
select_block(selected = "Species")
```

DivadNojnarg:
Fixed in 0.0.2 we can now do:

```r
library(blockr)
new_stack( # <1>
  data_block = new_dataset_block("penguins", "palmerpenguins"), # <2>
  filter_block = new_filter_block("sex", "female"), #<2>
  plot_block = new_ggplot_block("body_mass_g", "flipper_length_mm"), # <3>
  layer_block = new_geompoint_block("species", "species") # <4>
)
serve_stack(stack)
```

PR #350, opened 2024-04-05
wip: allow to pass in block calls in new_stack

DivadNojnarg:
@nbenn: added some code to be able to handle #78. I would have liked to not use `rlang`.  With this you can do:

```r
serve_stack(
  new_stack(
    data_block(selected = "iris"),
    select_block(columns = "Species"),
    title = "plop"
  )
)
```

The old case also works:

```r
serve_stack(new_stack(data_block, select_block))
```

Funnily, we probably still miss few checks to ensure people don't mix up things like (even though this technically works, I am not sure whether we should encourage this):

```r
serve_stack(
  new_stack(
    data_block(selected = "iris"),
    select_block,
    title = "plop"
  )
)
```

Note: will have to rename the branch...


R/stack.R
```diff
@@ -39,6 +42,39 @@ new_stack <- function(..., title = "Stack", name = rand_names()) {
...
+  UseMethod("init_stack_block")
+}
+
+#' @rdname init-stack-block
+#' @keywords internal
+init_stack_block.call <- function(x, position, ...) {
+  # For Nicolas: maybe you know how to get rid of rlang here ...
+  do.call(
+    rlang::call_name(x),
+    c(list(...), rlang::call_args(x), position = position)
```
DivadNojnarg, line 68: maybe data should not be passed in `...` and have a specific parameter.

DivadNojnarg:
As discussed with @nbenn, we'll implement a solution that basically allows to create blocks outside the stack. However this needs a rework of the infra we have: construction, initialisation and validation. See #110

PR #351, opened 2024-04-08
Download

JohnCoene:
This PR adds a download functionality.

The majority of the logic sits in `R/download.R`.

We create a few more generics to generate the download UI and server for blocks and stacks.

Note I have methods for the UI and the server but also for the `filename` and `content` arguments of the `downloadHandler`. This is so these can be reused at the stack level where the download button is set to download, not the stack, but the output of the last block.

```r
downloadHandler(filename, content)
```


JohnCoene:
This is ready for initial feedback.

R/download.R
```diff
@@ -0,0 +1,146 @@
...
+#' @export
+download_filename.plot_block <- function(x, object, ...) {
+  paste0(
+    attr(x, "name"),
+    ".png"
+  )
+}
+
+#' @export
+download.stack <- function(x, session, object, vals, ...) {
```
DivadNojnarg, line 127: download.stack only gets the last block data but I can also manually download data from the last block via the block download button. I don't see any added value from this method. Do we want instead to download multiple assets from the stack: data, plots?

DivadNojnarg: I would see benefits by making this method able get all the block data/plot instead of just the last block.

JohnCoene: I don't see it. This input is not intended for block or dashboard developers, it's for dashboard users who can only see the last block but would download all steps in-between? A dashboard developer can see the expanded stack and download individual blocks if necessary.

This would also make download likely fairly large as they would have to include everything, e.g.: a very large starting dataset. We may also have multiple transform blocks that do minor operations and essentially almost duplicate a lot of data.

JohnCoene: Thinking about this some more I should maybe add something to handle download of large files (zip?) in the event people want to download them, or maybe even completely prevent it?

How do we handle a user downloading a dataset of 2 mil+ rows?

JohnCoene: You can pull the latest on this branch and try again, sorry.

DivadNojnarg: Thanks. Working well.

DivadNojnarg (approved):
Thanks. LGTM.

R/download.R
```diff
@@ -0,0 +1,146 @@
...
+#' @inheritParams ui_fields
+#' @rdname download_ui
+#' @export
+download_ui <- function(x, ...) {
+  UseMethod("download_ui", x)
+}
+
+#' @rdname download_ui
+#' @export
+download_ui.default <- function(x, ns, inputs_hidden = FALSE, ...) {
```
nbenn, line 14: Why such a default method? Would you not rather have `download_ui(iris)` error to make the caller aware of the fact that this is not what it's meant for?

R/download.R
```diff
@@ -0,0 +1,146 @@
...
+  downloadLink(
+    outputId = id,
+    class = sprintf("cursor-pointer text-decoration-none stack-download %s", inputs_hidden),
+    iconDownload()
+  )
+}
+
+#' @rdname download_ui
+#' @export
+download_ui.data_block <- download_ui.transform_block
```
nbenn, line 44: What's the reason for not simply providing a `download_ui.block()` implementation and overriding this if necessary for a given block class? This is exactly what we have our inheritance structure for.

JohnCoene:
Thanks @DivadNojnarg, before merging, I will push something to try to mitigate potential issues from downloading large objects and let you know when it is done.

R/download.R
```diff
@@ -0,0 +1,146 @@
...
+}
+
+download_content <- function(x, object, file, ...) {
+  UseMethod("download_content", x)
+}
+
+#' @export
+download_content.plot_block <- function(x, object, file, ...) {
+  grDevices::png(filename = tempfile())
+  object()
```
nbenn, line 83: Why don't you just pass a value instead of a reactive? I'd rather we keep the shiny-specific data structures contained as much as possible.

R/download.R
JohnCoene: The idea was that if the block does not support download it does not generate the corresponding UI.

R/download.R
JohnCoene: It's the reason I mentioned prior I think.

e.g.: I create custom block that creates a plotly output, the default `download.block` method would fail. As far as I can tell, we can't really have a generic download that would work with all objects, unless I'm missing something.

R/download.R
nbenn: But then this should be handled in `download_ui.block()`, no?

R/download.R
JohnCoene: This needs to be reactive I believe, we can't use a reactive outside an observe which this would equate to. I think it's also to make it nicely work with the download at stack level which must be reactive because we fetch the last block.

JohnCoene:
- removed stack download
- no method on `default`, method on `block` instead
- added function to downsize object to download if too large @nbenn may I have your feedback on that?

https://github.com/blockr-org/blockr/blob/0232b8863ef57b2901d943dc264f726d9f4e71f2/R/download.R#L120

JohnCoene:
Can someone review this please?

nbenn (commented):
2 comments form my side
- re "downsampling": wouldn't it make more sense to error if the downloaded object is too large and leave it up to the user to further process such that things become more manageable (or allow to continue and download a large file anyways)?
- now that we no longer have the stack download, could we pass values instead of reactives in `download_content()`?
feel free to disregard either of the comments if it complicates things on your end unreasonably or if we're under time pressure to merge this.

JohnCoene:
I can't remove the reactive, it has to be one. If we pass the value returned by the reactive this will always download the data as it was at block initialisation, we need this to be reactive.

I've removed the downsizing of the data.

PR #354, opened 2024-04-12
Swap constructive for base serialize

### LINKED ISSUE #353: Redo complete object serialization
With the current approach using constructive we run into issues when serializing plot blocks. In order to fix this we can (for now) stick with the "complete object serialization" approach (#348) but use base R serialization infrastructure instead. Medium term we probably want to move towards a partial serialization approach to provide at least a bit of isolation of the code-base from serialized objects.

nbenn:
Fixes #353


JohnCoene:
@nbenn on how stacks are restored.

To recap, user creates a dashboard, this gets saved and restored.
Because of the previous issues with ggplot2 I currently use `save()` to "serialise" and `get(load())` to retrieve from the "serialised" file. Note that I remove the `result` attribute from the stacks because we are not allowed to save data, plus this could lead to very large amounts of data being stored.

I will try to find some time to switch what I have to this new PR soon. Hopefully that solves the issue with the serialisation.

I believe the challenges currently are not so much with the serialisation but the restoration of stacks in the dashboard.

The dashboard is saved (with `save()`) in a format resembling:

```r
list(
  tabs = list(), # data on the tabs (list of list where 1 list = 1 tab)
  workspace = blockr::get_workspace() |>
    ls() |>
    lapply(get, envir = ws) |>
    lapply(\(stack) {
      attr(stack, "result") <- NULL
      return(stack)
    })
)
```

So when I restore a dashboard I loop over the tabs, restore the tab, in each tab I have stored which stack it includes (tab stores stack `name` attribute) so I extract that from the workspace and restore it with `generate_ui(stack)` that I insert in the UI, and then run `generate_server()`. The code for this is [here](https://github.com/blockr-org/blockr.app/blob/master/R/restore.R#L91)

:white_check_mark: this works just fine for simple setups although you will notice a [`Sys.sleep(0.5)`](https://github.com/blockr-org/blockr.app/blob/ed15bde6440dd05f26ce2bd5186ce34febd354ae/R/restore.R#L119) before calling `generate_server()`. This should not really be the concern of blockr and is on my end to some extend.

A quick note on the above:

Shiny now uses [`processDeps()`](https://github.com/rstudio/shiny/blob/3edf9bfad8ad69a93f744a4f63447442d23e7c9a/R/html-deps.R#L45) R-side to pre-process any collection of `htmltools::tags`, this function essentially splits the content of htmltools into `deps` (dependencies, namely JS, CSS) and `html`, the content. This is then sent to the front-end via websocket using `session$sendCustomMessage` and rendered with `renderDependenciesAsync` __then__ `renderContentAsync`. This makes sure we render the content of any element __after__ its required dependencies have been injected into the DOM. This is what is done in masonry [here](https://github.com/blockr-org/masonry/blob/a08f81daa21e23cfa16a4958859788d1823c568c/srcjs/shiny.js#L62) to insert the stack in the UI when we restore the dashboard.

This adds some level of complexity because it runs asynchronously it is often the case that `generate_server()` runs __before__ the UI from `generate_ui()` is actually present in the UI and its `input`s registered with R. What's curious here though is that, as far as I know, this should not be a problem (rendering UI before server) but I've observed issues here with blockr stacks; hence the `Sys.sleep(0.5)`.

In any event we'll have to find a solution to this since `Sys.sleep()` is 1) not elegant and 2) sometimes 500 ms is not enough: it depends on the amount of content (~number of blocks) to restore.

:red_circle: where the restore definitely fails is with linked stacks.

When we restore linked stacks it's not just about the order of rendering the stack UI prior to calling its server. It's about calling the stacks in the correct order __and__ making sure we call stacks that depend on another __after__ the stack we depend on has been fully initialised server-side. e.g.: stack 1 joins with stack 2 so stack 1 depends on stack 2. Even if we call `generate_server` of stack 2 before we restore stack 1 it can fail because stack 1 has nothing on the `result` attribute yet (_remember that the `result` attribute has been removed from the stacks_). Note that this is my observation and understanding of the issues I observe.

As I see it the difficulties with restoring lie server-side. We could work something out where we call `generate_ui` on all stacks, place those UIs into the DOM, wait until they are rendered, then call `generate_server` on all of these (for us to do that, ideally, I would have the add-block PR merged, it greatly simplifies the code app side and makes this possible). But server-side we have the issues I have detailed prior to contend with: 1) restore in the correct order (stack dependencies) and 2) ensure this is done correctly (wait for server functions to be fully initialised).

@nbenn at least that is what I have observed and my perspective feel free to let me know what you think.

nbenn:
1. Remove result data from stacks
2. Look into #366

nbenn:
@DivadNojnarg this is ready for review.

One problem I ran into w/ shinytest2: somehow my snapshot-screenshots behave differently from yours. I had to re-generate those for the "restore-workspace" test and in doing so it asked to also update the other workspace-related snapshots. I could of course have skipped that, but this does not solve the underlying issue that between the two of use, things look differently. If you check out one such diff

https://github.com/blockr-org/blockr/pull/354/files#diff-af192cebe7bc6f2f996497023ed2cdd22ee6740abb2606bf0d4d16ac98f53334

can you tell what we have to do to make this look the same for both of us? Currently, buttons and inclusion of whitespace differs. Do I have an oder bslib installed maybe? Do we want to fix this in DESCRIPTION?

DivadNojnarg:
@nbenn That's the dependencies I have `bslib 0.7.0`, `shiny 1.8.1.1`:

```r
sessioninfo::session_info()
─ Session info ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.4.0 (2024-04-24)
 os       macOS Sonoma 14.5
 system   aarch64, darwin20
 ui       X11
 language (EN)
 collate  en_US.UTF-8
 ctype    en_US.UTF-8
 tz       Europe/Zurich
 date     2024-06-07
 pandoc   3.1.8 @ /usr/local/bin/pandoc

─ Packages ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ! package        * version    date (UTC) lib source
 P blockr         * 0.0.1.9000 2024-06-05 [?] local
   blockr.data      0.0.0.9000 2024-05-03 [1] Github (blockr-org/blockr.data@39096ad)
   brio             1.1.5      2024-04-24 [1] RSPM (R 4.4.0)
   bslib            0.7.0      2024-03-29 [1] RSPM (R 4.4.0)
   cachem           1.1.0      2024-05-16 [1] RSPM (R 4.4.0)
   cli              3.6.2      2023-12-11 [1] RSPM (R 4.4.0)
   colorspace       2.1-0      2023-01-23 [1] CRAN (R 4.4.0)
   constructive     0.3.0      2024-03-05 [1] RSPM (R 4.4.0)
   crosstalk        1.2.1      2023-11-23 [1] RSPM (R 4.4.0)
   desc             1.4.3      2023-12-10 [1] RSPM (R 4.4.0)
   devtools         2.4.5      2022-10-11 [1] RSPM (R 4.4.0)
   digest           0.6.35     2024-03-11 [1] RSPM (R 4.4.0)
   dplyr            1.1.4      2023-11-17 [1] RSPM (R 4.4.0)
   DT               0.33       2024-04-04 [1] RSPM (R 4.4.0)
   ellipsis         0.3.2      2021-04-29 [1] RSPM (R 4.4.0)
   fansi            1.0.6      2023-12-08 [1] RSPM (R 4.4.0)
   farver           2.1.1      2022-07-06 [1] CRAN (R 4.4.0)
   fastmap          1.2.0      2024-05-15 [1] RSPM (R 4.4.0)
   fontawesome      0.5.2      2023-08-19 [1] RSPM (R 4.4.0)
   forcats          1.0.0      2023-01-29 [1] RSPM (R 4.4.0)
   fs               1.6.4      2024-04-25 [1] RSPM (R 4.4.0)
   generics         0.1.3      2022-07-05 [1] RSPM (R 4.4.0)
   ggiraph        * 0.8.9      2024-02-24 [1] CRAN (R 4.4.0)
   ggplot2        * 3.5.1      2024-04-23 [1] CRAN (R 4.4.0)
   glue             1.7.0      2024-01-09 [1] RSPM (R 4.4.0)
   gtable           0.3.5      2024-04-22 [1] CRAN (R 4.4.0)
   haven            2.5.4      2023-11-30 [1] RSPM (R 4.4.0)
   hms              1.1.3      2023-03-21 [1] RSPM (R 4.4.0)
   htmltools        0.5.8.1    2024-04-04 [1] RSPM (R 4.4.0)
   htmlwidgets      1.6.4      2023-12-06 [1] RSPM (R 4.4.0)
   httpuv           1.6.15     2024-03-26 [1] RSPM (R 4.4.0)
   jquerylib        0.1.4      2021-04-26 [1] RSPM (R 4.4.0)
   jsonlite         1.8.8      2023-12-04 [1] CRAN (R 4.4.0)
   labeling         0.4.3      2023-08-29 [1] CRAN (R 4.4.0)
   later            1.3.2      2023-12-06 [1] RSPM (R 4.4.0)
   lifecycle        1.0.4      2023-11-07 [1] RSPM (R 4.4.0)
   magrittr         2.0.3      2022-03-30 [1] RSPM (R 4.4.0)
   memoise          2.0.1      2021-11-26 [1] RSPM (R 4.4.0)
   memuse           4.2-3      2023-01-24 [1] RSPM (R 4.4.0)
   mime             0.12       2021-09-28 [1] RSPM (R 4.4.0)
   miniUI           0.1.1.1    2018-05-18 [1] RSPM (R 4.4.0)
   munsell          0.5.1      2024-04-01 [1] CRAN (R 4.4.0)
   palmerpenguins   0.1.1      2022-08-15 [1] CRAN (R 4.4.0)
   pillar           1.9.0      2023-03-22 [1] RSPM (R 4.4.0)
   pkgbuild         1.4.4      2024-03-17 [1] RSPM (R 4.4.0)
   pkgconfig        2.0.3      2019-09-22 [1] RSPM (R 4.4.0)
   pkgload          1.3.4      2024-01-16 [1] RSPM (R 4.4.0)
   profvis          0.3.8      2023-05-02 [1] RSPM (R 4.4.0)
   promises         1.3.0      2024-04-05 [1] RSPM (R 4.4.0)
   purrr            1.0.2      2023-08-10 [1] RSPM (R 4.4.0)
   R6               2.5.1      2021-08-19 [1] RSPM (R 4.4.0)
   ragg             1.3.0      2024-03-13 [1] RSPM (R 4.4.0)
   Rcpp             1.0.12     2024-01-09 [1] RSPM (R 4.4.0)
   remotes          2.5.0      2024-03-17 [1] RSPM (R 4.4.0)
   rlang            1.1.3      2024-01-10 [1] CRAN (R 4.4.0)
   rprojroot        2.0.4      2023-11-05 [1] RSPM (R 4.4.0)
   rstudioapi       0.16.0     2024-03-24 [1] RSPM (R 4.4.0)
   sass             0.4.9      2024-03-15 [1] RSPM (R 4.4.0)
   scales           1.3.0      2023-11-28 [1] CRAN (R 4.4.0)
   sessioninfo      1.2.2      2021-12-06 [1] RSPM (R 4.4.0)
   shiny          * 1.8.1.1    2024-04-02 [1] RSPM (R 4.4.0)
   shinyAce         0.4.2      2022-05-06 [1] RSPM (R 4.4.0)
   shinyFiles       0.9.3      2022-08-19 [1] RSPM (R 4.4.0)
   stringi          1.8.3      2023-12-11 [1] RSPM (R 4.4.0)
   stringr          1.5.1      2023-11-14 [1] RSPM (R 4.4.0)
   systemfonts      1.0.6      2024-03-07 [1] RSPM (R 4.4.0)
   testthat       * 3.2.1.1    2024-04-14 [1] RSPM (R 4.4.0)
   textshaping      0.3.7      2023-10-09 [1] RSPM (R 4.4.0)
   tibble           3.2.1      2023-03-20 [1] RSPM (R 4.4.0)
   tidyselect       1.2.1      2024-03-11 [1] RSPM (R 4.4.0)
   urlchecker       1.0.1      2021-11-30 [1] RSPM (R 4.4.0)
   usethis          2.2.3      2024-02-19 [1] RSPM (R 4.4.0)
   utf8             1.2.4      2023-10-22 [1] RSPM (R 4.4.0)
   uuid             1.2-0      2024-01-14 [1] CRAN (R 4.4.0)
   vctrs            0.6.5      2023-12-01 [1] RSPM (R 4.4.0)
   withr            3.0.0      2024-01-16 [1] RSPM (R 4.4.0)
   xtable           1.8-4      2019-04-21 [1] RSPM (R 4.4.0)
   yaml             2.3.8      2023-12-11 [1] RSPM (R 4.4.0)

 [1] /Users/davidgranjon/Library/R/arm64/4.4/library
 [2] /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library
```

Screenshot 2024-06-07 at 10 44 51

nbenn:
Thanks @DivadNojnarg updating stuff did the trick.

DivadNojnarg (approved):
Thanks @nbenn! LGTM

PR #357, opened 2024-04-13
Turn off validation UI-side

JohnCoene:
This simply turns off validation UI-side.


DivadNojnarg (approved):
LGTM. Thanks @JohnCoene. I hope we can quickly get the new validation up and running.

PR #362, opened 2024-04-24
Use stack result when appending block

nbenn:
@JohnCoene this should fix the issue where we append a block to the end of the stack, as we there already have the cached stack result available.

If you want to insert a block at an arbitrary position, we could make it work by caching block level results, which we currently don't do and here we have to consider the memory/speed trade-off.

### DIFF
--- a/R/stack.R
+++ b/R/stack.R
@@ -178,10 +178,12 @@ add_block <- function(stack, block, position = NULL) {
   stopifnot(is_count(position))
 
   if (position == length(stack)) {
-
     data <- get_stack_result(stack)
+  } else {
+    data <- NULL
+  }
 
-  } else if (length(stack) > 0L) {
+  if (is.null(data) && length(stack) > 0L) {
 
     data <- evaluate_block(stack[[1L]])
 
--- a/R/stack.R
+++ b/R/stack.R
@@ -181,7 +181,7 @@ add_block <- function(stack, block, position = NULL) {
 
     data <- get_stack_result(stack)
 
-  } else {
+  } else if (length(stack) > 0L) {
 
     data <- evaluate_block(stack[[1L]])
 
--- a/R/stack.R
+++ b/R/stack.R
@@ -169,19 +169,26 @@ add_block <- function(stack, block, position = NULL) {
     position <- length(stack)
   }
 
-  log_debug("ADD BLOCK (position ", position + 1, ")")
-
   if (position < 1L) {
     position <- 1L
   }
 
-  # get data from the previous block
-  if (length(stack) == 1) {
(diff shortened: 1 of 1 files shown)


JohnCoene:
That seems to work fine, thank you! Can we merge?

JohnCoene:
Sorry, there's an issue somewhere, do not merge, I'm preparing a reprex

JohnCoene:
This fails.

```r
library(shiny)
library(blockr)

new_slow_data_block <- function(...) {
  expr <- bquote({
    Sys.sleep(5)
    data.frame(
      x = runif(100)
    )
  })

  new_block(
    fields = list(),
    expr = expr,
    ...,
    class = c("dataset_block", "data_block")
  )
}

slow_data_block <- function(...) {
  initialize_block(new_slow_data_block(...))
}

stack <- new_stack()

ui <- fluidPage(
  theme = bslib::bs_theme(5L),
  actionButton("add", "Add block"),
  generate_ui(stack)
)

  nb <- reactiveVal()

  observeEvent(input$add, {
    nb(
      list(
        position = 0L,
        block = new_slow_data_block
      )
    )
  })

  generate_server(stack, new_block = nb)
}

shinyApp(ui, server)
```

nbenn:
@JohnCoene could you be a bit more specific? In what way exactly does this fail and how is this related to the change proposed in this PR?

JohnCoene:
Starting from an empty stack and adding a first block via `new_block` reactive crashes the application. It is related to this PR as it works on `main`.

JohnCoene:
Sorry my bad, I see it now. If I can reproduce what I encountered I will create another reprex

JohnCoene:
The issue in the example is caused by passing a constructor rather than a `block_descr` object (because the slow block was not registered).

This is something that works on main but not on this branch: from an empty stack add a data block.

```r
library(shiny)
library(blockr)

stack <- new_stack()

ui <- fluidPage(
  theme = bslib::bs_theme(5L),
  actionButton("add", "Add block"),
  generate_ui(stack)
)

  nb <- reactiveVal()

  observeEvent(input$add, {
    nb(
      list(
        position = 0L,
        block = available_blocks()$dataset_block
      )
    )
  })

  generate_server(stack, new_block = nb)
}

shinyApp(ui, server)
```

DivadNojnarg (approved):
Only merge when fixed and confirmed by @JohnCoene.

nbenn:
@JohnCoene yes, that was a bug I introduced. Thanks for pointing it out. Should be fixed now.

JohnCoene:
Seems to work, thanks!

PR #365, opened 2024-04-29
Fix workspace without stack issue

### LINKED ISSUE #364: UI issue: Inserted stack don't show in workspace with initially 0 stacks
Because of a current condition in the code:

```r
if (length(stacks) > 0) {
    stack_ui <- div(
      class = "d-flex stacks flex-wrap",
      lapply(seq_along(stacks), \(i) {
        div(
          class = "flex-grow-1 stack-col m-1",
          id = sprintf("%sStackCol", names(stacks)[i]),
          generate_ui(stacks[[i]], ns(names(stacks)[i]))
        )
      })
    )
  }
```
To insert a stack the server needs a piece of UI having the `.stack` class which isn't the case with a workspace initialized with 0 stacks.

DivadNojnarg:
Fix #364.

Could you double check @JohnCoene if this is working correctly on your side?

Also, the example should start without any stack so that we can use shinytest2 and drive the app test from an empty workspace to avoid missing this check (adding stack to empty workspace).


DivadNojnarg:
TO DO: add some shinytest2 checks

JohnCoene:
Looks good

DivadNojnarg:
Good for quick review @nbenn. I took this opportunity to also include a simple shinytest2 script. There are some issues with shinytest2 and deeply nested reactive values like we have. Hence, I had to look at the very high level elements like the number of stacks in the workspace + other relevant input elements.

I'll do more testing in another PR for stacks (where I'll test to add/remove blocks, ...) and blocks (?). Side note: at the workspace level, with our current server function, I am not able to control how a newly added stack is named since no option is proposed, that's why it is better to split the tests (at the stack level, we can control the stack namespace).

nbenn (approved):
Lgtm, thanks @DivadNojnarg!

PR #367, opened 2024-04-30
Enable block ctors with default args

nbenn:
Notable changes/issues
- I renamed the data_block() constructor to be consistent with our naming scheme new_*class*()
- When running tests locally, some snapshots are removed; probably not what we want


R/blocks.R
```diff
@@ -282,12 +285,9 @@ result_block <- function(...) {
 #' @param filter_fun Default filter fun for the expression.
 #' @rdname filter_block
 #' @export
-new_filter_block <- function(
-    data,
-    columns = colnames(data)[1L],
```
nbenn, line 287: Changes to such default arg values is the core of this PR @DivadNojnarg, in order to remove dependencies on `data` in block constructors.

tests/testthat/test-block.R
```diff
@@ -271,10 +271,17 @@ test_that("xpt parser block", {
...
     tagList(
       span("blockr", class = "badge bg-light"),
       "Data"
     )
   )
 })
+
+test_that("blocks can be constructed with default args", {
+
+  for (block in names(available_blocks())) {
```
nbenn, line 284: @DivadNojnarg I added a test to go through all constructors and call them with no `data` arg and this is where I tripped over the inconsistency of the `data_block()` constructor naming. Something I have been wanting to change anyways for some time now.

nbenn:
@DivadNojnarg The PR removes all snaps in `mac-4.3/workspace`, probably by mistake, right? I'm running R 4.2, so that might be the root cause?

DivadNojnarg:
> @DivadNojnarg The PR removes all snaps in `mac-4.3/workspace`, probably by mistake, right? I'm running R 4.2, so that might be the root cause?

Yes, a new folder is created per OS-R version. Did you run the test locally before pushing? 

The package must also be installed before running the test (locally). On GA, 
> `devtools::install(); testthat::test_local()` or by running R CMD check on your package

DivadNojnarg (commented):
Thanks @nbenn

I also updated the shinylive codes (will only work when we merge to main and update the wasm CRAN repo). Check with @JohnCoene if this is not going to cause issue during the upcoming demo.

---
### Regarding the test issue (sorry if this comment is a bit long compared to the rest)

I don't get why you don't have a folder corresponding to your R version (should be mac-4.2) as I have below when switching to R4.2.3:

Screenshot 2024-05-01 at 16 16 25

Did the test went smoothly for you locally? In general snapshots get deleted because of an error preventing them to get reach in the testthat file. 

To simplify things, I removed the `platform_variant`, as I don't expect R4.5 to break blockr, but rather us doing a typo. Doing so, if I remove the `add stack` button by mistake, the test fails:

```shell
Error (test-workspace.R:117:3): workspace demo works
Error in `app_find_node_id(self, private, input = input, output = output, 
    selector = selector)`: Cannot find HTML element with selector #myworkspace-add_stack
Backtrace:
    ▆
 1. └─app$click(selector = "#myworkspace-add_stack") at test-workspace.R:117:3
 2.   └─shinytest2:::app_click(...)
 3.     └─shinytest2:::app_find_node_id(self, private, input = input, output = output, selector = selector)
 4.       └─shinytest2:::app_abort(...)
 5.         └─rlang::abort(..., app = self, call = call)

[ FAIL 2 | WARN 2 | SKIP 0 | PASS 342 ]
```
Note that because the above test failed after the first step, only the first snapshot is kept (the 3 other are removed because not used). I guess this is enough for us in term of strategy. 

---

Happy to discuss about this.

DivadNojnarg:
@nbenn Quick question: does this PR intend to also support creating block outside the stack and then instantiate a stack? From what I see I don't think so, but just want to double check with you. If not, then we'll need another PR to be able to do this, which currently fails since what we pass to the stack is not a constructor but a block object:

```r
blk <- dataset_block(name = "plop")

stacks <- lapply(
    1:3, 
    \(x) new_stack(blk, name = sprintf("stack%s", x))
)

Error in do.call(ctors[[1L]], list(position = 1)) : 
  'what' must be a function or character string
```

This relates to what I was checking in #350, except that we don't need to pass a call anymore, given that the block is created outside the stack. This is also a go opportunity to get rid of the `position` attribute.

DivadNojnarg:
From `set_workspace_stacks` code, we only check `!is.null(nms)` when `length(stacks) > 0`. The demo starts with an empty workspace, so we shouldn't enter the above condition:

```r
library(blockr)
do.call(set_workspace, args = list(title = "My workspace"))
```

I don't get the error on my side. Do you install the package before running the test locally?

DivadNojnarg:
Hidden somewhere in the shinytest2 pkgdown doc 📝:

> [shinytest2](https://rstudio.github.io/shinytest2/) requires that your package to be installed when testing. [testthat::test_local()](https://testthat.r-lib.org/reference/test_package.html) (and related wrappers) eventually call [pkgload::load_all()](https://pkgload.r-lib.org/reference/load_all.html) to temporarily source the local R package. You can use [test_local()](https://testthat.r-lib.org/reference/test_package.html) to test non-[shinytest2](https://rstudio.github.io/shinytest2/) tests, but you will need to install your R package to safely execute your [shinytest2](https://rstudio.github.io/shinytest2/) tests. If not installed, it will create a confusing situation where your [shinytest2](https://rstudio.github.io/shinytest2/) tests are running on a different version of your R package (whichever was last installed), than the rest of your tests (the current source).

The CICD pipeline installs it so we don't have issue there but locally, you have to run it manually.

DivadNojnarg (approved):
LGTM thanks

PR #368, opened 2024-05-01
More testing with shinytest2

### LINKED ISSUE #170: More coverage for shiny server logic
Some of the shiny server testing is part of https://github.com/blockr-org/blockr/pull/119. However, https://github.com/blockr-org/blockr/pull/119 has become irrelevant for other reasons and I'll just extract the tests from this PR in another PR. Will require #121 to be merged before.

DivadNojnarg:
Will fix #170

- [x] stack level.
- [x] block level.
- [x] restore workspace.


tests/testthat/_snaps/stack/stack-app-006.json
DivadNojnarg, line 1: @nbenn On this snapshot, I would expect the stack to be empty when it is removed. That said, it's not super important as we only care about blocks within the stack.

tests/testthat/test-stack.R
```diff
@@ -36,13 +35,88 @@ test_that("stacks", {
...
+  # Add a block
+  app$click(selector = ".stack-add-block")
+  app$set_inputs("mystack-selected_block" = "dataset_block")
+  app$click(selector = "#mystack-add")
+
+  app$expect_values(
+    input = stack_inputs,
+    export = stack_exports,
+    screenshot_args = list(
+      delay = 1
```
DivadNojnarg, line 79: delay required because of the animation. On a laggy system, that might not be enough but having too much delay would cause slow down the testing pipeline.

DivadNojnarg:
@nbenn I think we're good for review. The critical thing is also to capture plotting issues which I do by using `vdiffr::expect_doppelganger` on a dedicated export element `res` (as ggplot2 obj can't be serialized as json, I can't just check the output element provided by shinytest2 with `expect_values` or `get_values` and need to send the plot object in the export list to be processed by `{vdiffr}`):

```r
test_plot <- function(i) {
    message("TESTING PLOTS")
    plot_obj <- app$get_values(
      export = sprintf("my_stack-block_%s-res", i)
    )
    # Verify `plot_obj()` is consistent
    vdiffr::expect_doppelganger(
      sprintf("check-plot-%s", i),
      plot_obj
    )
  }

lapply(..., test_plot)
```

 Below I deliberately set the plot to be NULL to see whether the test would fail (it should) and the output is consistent, allowing to preview the changes with `{diffviewer}` and accept/ignore.

Screenshot 2024-05-03 at 09 45 13

@JohnCoene Do you see any critical missing piece here?

nbenn (approved):
LGTM, thanks!

PR #373, opened 2024-05-10
Validation rework

nbenn:
Remove the dual constructor setup: the single `new_*()` block constructor does
not initialize via `initialize_block()`.


DivadNojnarg (commented):
@nbenn 
Compared to the main branch, the validation (server side) is causing issues and I can't run any demo app like this one:

```r
library(blockr)

stack <- new_stack()
id <- "mystack"

shiny::shinyApp(
  ui = bslib::page_fluid(
    generate_ui(stack, id)
  ),
    stack <- generate_server(stack, id)
  }
)
```
Screenshot 2024-05-13 at 07 59 38

as the first block appears invalid, downstream block don't show any data or any choice in the fields.

If this is a temporary broken state and is fixed by another PR, please ignore my comment. Otherwise, this needs to be fixed.

R/blocks.R
```diff
@@ -4,40 +4,27 @@
...
-    list(data = substitute(dat))
-  )
+    gsub("\\s+\\(.+\\)$", "", datasets)
+  }
 
   new_block(
-    fields = fields,
-    expr = expr,
+    fields = list(
+      package = new_hidden_field(package, type = "name"),
```
DivadNojnarg, line 23: @nbenn `hidden_field` don't play well with the current validation. The error we see is coming from here. At this point, either we 1) remove everything or 2) we temporarily patch it. We could opt for 1 if this does not break everything else, and would also prepare the work for the new validation implementation.

R/blocks.R
```diff
@@ -4,40 +4,27 @@
...
+      dataset = new_select_field(selected, list_datasets,  type = "name",
```
DivadNojnarg, line 24: By default, the first newly selected dataset is not rectangular (AirPassengers) which triggers error.

nbenn: Yes, sorry, this is a regression. We had fixed this at one point and this fix was lost. Re-added it now.

R/blocks.R
nbenn: The problem here was that hidden fields are excluded based on field name instead of class attributes. I tried fixing things by returning `character()` for the hidden field-specific `input_ids()`. Not sure whether this causes other issues, but it might be more correct from a conceptual pov compared to what we had previously.

DivadNojnarg (approved):
I confirm the hidden field issue went away. I checked at the stack and workspace level with all the included blocks.

DivadNojnarg:
@nbenn I did further basic testing here from the stack demo code from this branch and observed a crash which I don't get from the main branch.

## Error

```shell
[ERROR][2024-05-17 15:48:14.560133] Error in `[.data.frame`(data, , cols, drop = FALSE) : 
  undefined columns selected

Warning: Error in UseMethod: no applicable method for 'value<-' applied to an object of class "NULL"
  83: value<- [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/field-core.R#190]
  82: update_sub_fields [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/field-core.R#234]
  80: validate_field.list_field [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/fields.R#373]
  78: update_field.field [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/field-core.R#112]
  76: update_blk [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/server.R#68]
  75: observe [/Users/davidgranjon/david/Cynkra/BMS/blockr/R/server.R#154]
  74: <observer>
   3: runApp
   2: print.shiny.appobj
   1: <Anonymous>
```

## Code used

```r
library(blockr)

stack <- new_stack()
id <- "mystack"

shiny::shinyApp(
  ui = bslib::page_fluid(
    generate_ui(stack, id)
  ),
    stack <- generate_server(stack, id)
  }
)
```

## Steps to reproduce 

1. Add dataset block (with BOD dataset)
2. Add select block.
3. Add filter block.
4. Remove default selected column in select block and change by the other column.
5. Manually update filter.
6. Crash.

nbenn:
Changes to how blocks are implemented (WIP):

- `data` is no longer passed to blocks
- we no longer have a dual constructor set-up, but only one per object (also applies to fields)
- functions which are passed as field components should behave as pure functions (and not rely on any closure-like properties they do have)

DivadNojnarg:
Still few issues:

- Remove any entry from a list_field crashes the block: filter block, select a column, then remove it -> crash
- Issue with wrong validation message shown with: dataset block + select block + filter block -> Currently checking.

nbenn:
@DivadNojnarg fixed the app crash on list field removal. That was an oversight on my end. You can now trigger an infinite loop w/ this.

Re block server refactoring: I'm with you on that. It's something we should do. If possible I'd do it in a separate PR though. This one already is huge and it would be nice to be able to merge this w/o rewriting the block servers. What do you think?

DivadNojnarg:
@nbenn Yes I agree, the server function refactoring would be too important for this PR. The things I would fix are the tests and if possible try to understand what happened to the plot layer elements.

DivadNojnarg:
Regarding the plot layer issue, this is coming from the `data`. In main `data` contains the plot object and now it contains the `result` (rectangular data), so what happens now is that we do `iris + geom_point(...)`.

JohnCoene:
I was trying to re-create the old faithful with blockr for the UseR2024 presentation (see the example below).

Perhaps I'm missing something but changing the dataset block does not update the select input in the block below.

```r
library(blockr)

plot_histogram <- function(
  data, 
  column,
  breaks = 20, 
  obs = FALSE, 
  density = FALSE, 
  bw_adjust = 1,
  color = "blue"
){
  values <- data[[column]]
  hist(
    values,
    probability = TRUE,
    breaks = breaks
  )

  if (obs) {
    rug(values)
  }

  if (density) {
    dens <- density(
      values,
      adjust = bw_adjust
    )

    lines(dens, col = color)
  }
}

new_histogram_block <- function(...){
  first_col <- \(data) data |> dplyr::select_if(is.numeric) |> colnames() |> (\(x) x[1])()
  columns <- \(data) data |> dplyr::select_if(is.numeric) |> colnames()

  new_block(
    name = "Histogram",
    expr = quote(plot_histogram(.(column))),
    class = c("histogram_block", "plot_block"),
    fields = list(
      column = new_select_field(first_col, columns)
    ),
    ...
  )
}

stack <- new_stack(
  new_dataset_block,
  new_histogram_block
)

serve_stack(stack)
```

issue

DivadNojnarg:
@karmatarap. Seeing this warning in the tests:

```r
Warning (test-block.R:346:3): sas parser block
`write_sas()` was deprecated in haven 2.5.2.
i Please use `write_xpt()` instead.
```

Should we remove the sas blocks and leave xpt only?

DivadNojnarg:
@nbenn We now have "only" 4 failing test. I don't get what's wrong with the 2 first ones (can't reproduce, works well for me). The 2 others relate to `value` which does not seem to work as expected.

```r
── Failure ('test-field.R:127:3'): upload field ────────────────────────────────
`validate_field(field)` did not throw the expected error.
── Failure ('test-field.R:149:3'): filesbrowser field ──────────────────────────
`validate_field(field)` did not throw the expected error.
── Failure ('test-field.R:189:3'): list field ──────────────────────────────────
value(field) (`actual`) not identical to list("bar") (`expected`).

`actual[[1]]`:   "foo"
`expected[[1]]`: "bar"
── Failure ('test-field.R:201:3'): list field ──────────────────────────────────
value(field) (`actual`) not identical to list("bar", 2) (`expected`).

`actual[[1]]`:   "foo"
`expected[[1]]`: "bar"

  `actual[[2]]`: 3
`expected[[2]]`: 2
```

DivadNojnarg:
@nbenn if we can merge asap, as I have to update the useR slides (with shinylive demo) for Monday next week.

nbenn:
@DivadNojnarg @JohnCoene Many thanks for your contributions here!

PR #374, opened 2024-05-10
Remove dual field constructors

nbenn:


tests/testthat/iris.csv
DivadNojnarg, line 1: @nbenn Is this new file needed/expected?

DivadNojnarg (commented):
TBD: update all the shinylive and Rmd code examples to comply with this new change.

Same as for #373, I can't test examples as the hidden field from data_block is causing a validation issue.

tests/testthat/iris.csv
nbenn: Nope, my bad, sorry. I had experimented with [`withr::local_file()`](https://withr.r-lib.org/reference/with_file.html), but but somehow did not properly manage the mechanics of the API there. Replaced all with [`local_tempfile()`](https://withr.r-lib.org/reference/with_tempfile.html) and removed leftovers.

DivadNojnarg:
I am currently checking, the plot layer vignette which needs to be updated since we pass data as parameter and we still have the 2 block constructors. We should have something like:

```r
new_ggplot_block <- function(...) {

  new_block(
    fields = list(
      x = new_select_field("wt", data_cols, type = "name"),
      y = new_select_field("qsec", data_cols, type = "name"),
      color = new_select_field("disp", data_cols, type = "name")
    ),
    expr = quote(
      ggplot(mapping = aes(x = .(x), y = .(y), color = .(color)))
    ),
    ...
  )
}
```

This applies to all vignettes still using our old approach.

PR #376, opened 2024-05-21
No-fix validation

nbenn:
Instead of trying to resolve inconsistencies that might arise from changes upstream in a stack by resetting incompatible values to defaults, we now fail validation and require user feedback to resolve issues.


DivadNojnarg:
@nbenn We probably should not call `validate_field` inside `initialize_field`?
See https://github.com/blockr-org/blockr/pull/376/files#diff-8546c791c7ba260a074d6a5270c42f3e71c421dd507effb82059def8f8316d53R63.

This currently returns NULL.

nbenn:
@DivadNojnarg agreed. I modified `initialize_field()` and `update_field()` a bit. Not sure if this is what we want exactly, but it should be better than what we had.

R/field-core.R
```diff
@@ -90,21 +88,23 @@ update_field <- function(x, new, env = list()) {
...
+  validate_field(x)
 
-  if (is_initialized(res)) {
-    return(res)
+  if (is_initialized(x)) {
+    return(x)
   }
 
-  eval_set_field_value(res, env)
+  eval_set_field_value(x, env)
```
DivadNojnarg, line 107: @nbenn 

Reading detail:  from semantic POV, can't we call `initialize_field`? This does exactly the same in the end (unless someone defines a specific initialize method for given specific field, very unlikely though)

Then we can read: if field is initialized --> return field, otherwise initialize it.

nbenn: I completely agree that our current default `update_field()` is a bit of a mess. Some thoughts:

- `eval_set_field_value()` simply evaluates field components that are functions
- we have to call (something along the lines of) `eval_set_field_value()` after every update (we can make this smarter to only fire if relevant things have changed, but for now, we just do it on every change)
- `initialize_field()` is something that sounds like it should only run once after a field object has been instantiated (maybe we can remove this alltogethe, not sure currently)

A cleaned up implementation of `update_field()` could look something like

```r
update_field <- function(x, new, env = list()) {
  UseMethod("update_field", x)
}

update_field.field <- function(x, new, env = list()) {

  if (length(new)) {
    value(x) <- new
  }

  res <- eval_set_field_value(x, env)

  validate_field(res)

  res
}

update_field.hidden_field <- function(x, new, env = list()) {

  res <- eval_set_field_value(x, env)

  validate_field(res)

  res
}
```

How does that look to you @DivadNojnarg?

nbenn: Another thing we should look into: currently, `eval_set_field_value()` is run even if some required inputs are not yet available. This was the reason that such functions specified in block definitions need to have "safety-nets" built in, for example

https://github.com/blockr-org/blockr/blob/238cecdd989353196067126556a9dc67089ffce0/R/blocks.R#L63-L65

This makes block implementation unnecessarily hard and we should try to handle as much as this on a systematic level as possible.

DivadNojnarg: This looks cleaner. 

About `hidden_field`, I had some issues with the `update_blk` function that somehow shunts the `update_fields` generic. What happens is that the field has correct initial state but because there's no UI input, it's value is set to nothing when `update_blk` is called. I had to include some check (see my comment) so the update does not happen

```r
update_blk <- function(b, value, is_srv, input, data) {
  for (field in names(b)) {
    if (field %in% names(is_srv)[is_srv]) {
      b[[field]] <- update_field(b[[field]], value[[field]])
    } else {
      env <- c(
        list(data = data),
        value[-which(names(value) == field)]
      )
      # Don't update hidden field because they don't have UI components
      # so their value would be set to NULL
      if (!inherits(b[[field]], "hidden_field")) {
        b[[field]] <- update_field(b[[field]], value[[field]], env)
      }
    }
  }
  b
}
```

DivadNojnarg: @nbenn What do you think of including `initialize_field` within `update_field`. Update field, is kind of re-initializing the field with a new value.

```r
update_field.field <- function(x, new, env = list()) {

  if (length(new)) {
    value(x) <- new
  }

  initialize_field(x)
}
```

nbenn: @DivadNojnarg do you know why `update_field.hidden_field()` is not picked up? The issue you describe is precisely why we override the `field`-level implementation.

nbenn: @DivadNojnarg the reason why I did not use `initialize_field()` is b/c in an update setting it feels a bit wrong (semantically). The field has already been initialized after all. It might be re-initialized or something. But that is not actually what is happening. It's more being *evaluated*. Maybe we could come up with a better name for `eval_set_field_value()` and expose that directly as generic? Maybe something like `evaluate_field()` or have an `evaluate()` generic that can be dispatched on both `block` and `field` (we currently already have `evaluate_block()` which then would go away).

DivadNojnarg: > do you know why update_field.hidden_field() is not picked up?

Nope. Currrently checking.

`evaluate_field` would be more meaningful that `eval_set_field_value`.

DivadNojnarg: Slowly getting there on the server side...

Screenshot 2024-05-22 at 11 40 59

DivadNojnarg:
@nbenn. One remark about `validation_failure`: calling `rlang::abort` causes the app to crash. It should probably be handled differently, with `rlang::warn` as I did in [981f306](https://github.com/blockr-org/blockr/pull/376/commits/981f3069f2ccaa2e432016d7666e6aa06ddf68fe). Then on the shiny side, `req` ensures we stop the evaluation if the validation failed. Finally, the message is extracted and sent to the UI.

DivadNojnarg:
@nbenn update on where I am.

Currently, in the block server function, we execute the validation which in turns allows to do the computation but only for the current block. 

## Problem:

An issue that arises when having multiple blocks: assuming a stack with a data block and a select block, both valid. Empty the field from the data block, block 1 becomes invalid but somehow, the block2 server functions is triggered again and block2 appears valid because it still has the old values from the previous data. This fails because the columns don't exist in the new expression.

A piece of info is missing at the stack level.

## Possible solution: 

From the stack level, we should probably pass the validation state of the previous block and add a check at the top level of the block server function: `req(previous_block_valid)` so that this server function does not go further.

DivadNojnarg:
@nbenn Commit [e05ff8d](https://github.com/blockr-org/blockr/pull/376/commits/e05ff8d74a8a61f825b3e380d6786db3515790bd) fixes the error issue. No crash anymore. Now I need to solve [this](https://github.com/blockr-org/blockr/pull/376#issuecomment-2124446531).

R/server.R
```diff
@@ -237,7 +249,10 @@ generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "pl
       return(
         list(
           block = blk,
-          data = out_dat
+          data = out_dat,
+          # Needed by the stack to block
+          # computations for the next block
+          is_valid = reactive(is_valid$block)
```
DivadNojnarg, line 255: @nbenn This is the core change of this commit: the current block has to return its validation status for the next one. The stack is responsible for dispatching it to the next block.

R/server.R
```diff
@@ -648,7 +663,16 @@ init_block <- function(i, vals) {
       # Data from previous block
       vals$blocks[[i - 1]]$data
     },
-    id = attr(vals$stack[[i]], "name")
+    id = attr(vals$stack[[i]], "name"),
+    # Extract the state of the previous block
+    # to pass it to the next one. This is needed
+    # within the next block server function
+    # to reset calculations if required.
+    is_prev_valid = if (i == 1) {
```
DivadNojnarg, line 671: Dispatch by the stack ...

R/server.R
```diff
@@ -133,8 +134,17 @@ generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "pl
           generate_server(x_srv[[name]])(name, init = l_init[[name]], data = in_dat)
       }
 
+      # When the previous block changes data or validation
```
DivadNojnarg, line 137: Third change: whenever data or the previous block changes, we reset the current block validation status, so computation don't run by accident.

R/server.R
```diff
@@ -82,13 +82,14 @@ update_ui <- function(b, is_srv, session, l_init) {
   }
 }
 
-generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "plot")) {
+generate_server_block <- function(x, in_dat = NULL, id, display = c("table", "plot"), is_prev_valid) {
```
DivadNojnarg, line 85: @nbenn We both agree that the block server logic should be rewritten. Out of scope of this PR.

DivadNojnarg:
@nbenn That's where we currently are. Resetting the data/changing dataset seems to work for basic blocks.

https://github.com/blockr-org/blockr/assets/18291543/fe0a17ed-079a-472a-b53f-843608d13b58

Few issues:

- Selecting multiple values in the select block triggers a validation error. I imagine multiple values are not handled properly.
- The filter block seems broken (the nested field related to the column does not show).

DivadNojnarg:
Digging in the `list_field` issue, when I create a filter block and putting a browser within `ui_update.list_field`, I don't see any sub fields:

```r
x
$value
NULL

$sub_fields
character(0)

[[3]]
function(data, columns) {
    determine_field <- function(x) {
      switch(class(x),
        factor = new_select_field,
        numeric = new_range_field,
        new_string_field
      )
    }

    field_args <- function(x) {
      switch(class(x),
        factor = list(levels(x)[1L], choices = levels(x)),
        numeric = list(range(x), min = min(x), max = max(x)),
        list()
      )
    }

    cols <- data[, columns, drop = FALSE]

    ctor <- lapply(cols, determine_field)
    args <- lapply(cols, field_args)

    Map(do.call, ctor, args)
  }
<bytecode: 0x11f2706d0>
<environment: 0x10ce1b440>

attr(,"type")
[1] "literal"
attr(,"class")
[1] "list_field" "field"     
attr(,"title")
[1] "Value"
attr(,"descr")
[1] ""
attr(,"status")
[1] "active"
attr(,"exclude")
[1] FALSE
```

which means, nothing can be inserted.

JohnCoene:
Apologies if it is not the best place for comments, let me know if there is a better place.

- Should we export `validation_failure`? I know users can call `rlang::abort()` themselves but perhaps this would make it clearer as this is what we use internally + it would enable changing things in that function in the future? 
- Maybe picky of me but this PR has blocks following this pattern `*_block` rather than `new_*_block` which we used before, should we not do the opposite (just talking about function naming) as this would avoid breaking existing use of blockr

This is the example I have in #378 

```r
my_block <- function(...){
  all_cols <- function(data) colnames(data)

  first_col <- \(data) colnames(data)[1]
  second_col <- \(data) colnames(data)[2]

  fields <- list(
    x_var = new_select_field(first_col, all_cols),
    y_var = new_select_field(second_col, all_cols)
  )

  new_block(
    fields = fields,
    expr = quote({
      ggplot2::ggplot(
        data,
        mapping = ggplot2::aes_string(
          x = .(x_var),
          y = .(y_var)
        )
      ) +
        ggplot2::geom_point() +
        ggplot2::geom_line()
    }),
    ...,
    class = "my_block"
  )
}

.S3method("server_output", "my_block", function(x, result, output){
  renderPlot(result())
})

.S3method("uiOutputBlock", "my_block", function(x, ns){
  plotOutput(ns("res"))
})

# we have to implement this
.S3method("generate_server", "my_block", function(...){
  blockr:::generate_server_block(...)
})

stack <-  new_stack(
  new_dataset_block,
  my_block
)

ui <- fluidPage(
  theme = bslib::bs_theme(5L),
  generate_ui(stack)
)

  generate_server(stack)
}

shinyApp(ui, server)
```

In "My block" I cannot change any of the inputs, the inputs reset, perhaps I'm not doing this right, or it may be linked to one of the issues pointed out by @DivadNojnarg?

issue

DivadNojnarg:
@nbenn, after digging a bit more, I noticed that `get_sub_fields(x)` always return NULL, which is an issue as it is used within `value<-.list_field`:

```r
`value<-.list_field` <- function(x, name = "value", value) {
  set_sub_fields(x, Map(`value<-`, get_sub_fields(x), name, value))
}
```

This means, `Map(`value<-`, get_sub_fields(x), name, value)` does not yield the expected output, so is `set_sub_fields`.

DivadNojnarg:
@nbenn regarding the select input validation issue, it is coming from `is_string`:

```r
is_string <- function(x) {
  is.character(x) && is_scalar(x)
}
```

When passed multiple values, `is_scalar` is FALSE. 

The select validation is:

```r
validate_field.select_field <- function(x) {
  val <- value(x)
  validate_string(val)
...
}
```

Initial state, we have `character(0)` which fails the validation, as expected. Then, as soon as we select a value from the select input, we'd have to apply `validate_string` to any element (side note:  not necessary in Shiny's context as the selectInput does not allow to pass other things than strings in the UI and returns strings):

This would lead to something like:

```r
if (length(val) <= 1) {
    validate_string(val)
  } else {
   # useless in Shiny' context
    lapply(val, validate_string)
  }
```

The `else` being "useless" as explained before. Maybe this is still relevant as we plan to initialise blocks outside the stack (outside shiny' context).

nbenn:
@JohnCoene Yes, `validation_failure()` should absolutely be an exported function and this is what should be used in block validators. Sorry, this was an oversight on my end and should be fixed now.

nbenn:
@DivadNojnarg fixed the select field validation ([3f5cc7a](https://github.com/blockr-org/blockr/pull/376/commits/3f5cc7ae24ec6b3ce8e50090186ed581fe92c5a0)). Thanks for reporting & debugging.

nbenn:
@DivadNojnarg one problem with the list field `value()` and `value<-()` implementations was that what I had was only correct when manipulating the "value" component and not any other (such as the "sub_fields" component).

Still not quite there yet, as changes are currently being overwritten. Looking into that now.

nbenn:
@DivadNojnarg I believe the list field issue is now resolved. The `value<-()` replacement function is super ugly however. Currently the mechanics for these recursive fields is unsatisfyingly complex. We might want to re-think that (but not as part of this PR as unrelated).

nbenn:
@JohnCoene thanks for pointing out some issues you ran into.

- The `new_submit_field()` was not added as part of this PR. This is definitely something I want to change, but it is unrelated to this PR: the field is auto-added based on presence of a `submit_block` class attribute and it currently only works in such blocks (due to how it is integrated in the block server fun). Will tackle this separately.
- The filebrowser field works as part of the filebrowser block (for me at least and with a later state of the code that is). Did your issue maybe come up as part of another block?
- The range field works (as part of the filter block) for me with the current code. What was your context when you ran into an issue?

JohnCoene:
I see these issues in this block for instance.

```r
test_block <- function(...){
  first <- \() letters[1]
  all <- \() letters

  fields <- list(
    text = new_string_field("default"),
    select = new_select_field(first, all),
    bool = new_switch_field(TRUE),
    numeric = new_numeric_field(1L, min = 0, max = 10L),
    submit = new_submit_field(),
    browser = new_filesbrowser_field(),
    slider = new_range_field(1, 0, 10),
    range = new_range_field(c(1,2), 0, 10)
  )

  new_block(
    fields = fields,
    expr = quote(iris),
    ...,
    class = c("data_block", "my_block")
  )
}
```

PR #379, opened 2024-05-30
Validation UI

JohnCoene:
It's difficult to know where the errors are coming from since the messages are just displayed in a list at the top of the list: the user has to guess which input has an issue.

image

This small PR branches off 110-4-no-fix and makes small changes to display the error messages below their corresponding fields.

image


nbenn (commented):
Lgtm, thanks!

nbenn:
@JohnCoene no objections from my side, I think this is a good suggestion. @DivadNojnarg would you take a quick look? This affects more your change than mine.

DivadNojnarg (approved):
All good for me. Thanks @JohnCoene!

PR #388, opened 2024-06-25
Rendered block loading not disappearing

JohnCoene:
This a (probably temp) fix to the loading issue. 

On my end the loading icon sometimes does not go away.

@DivadNojnarg it's not a super elegant solution but it's to do with the (AFAIK) new Shiny's renderAsync function, it's impossible to tell when a piece of UI is rendered.


DivadNojnarg (commented):
Thanks @JohnCoene. Looks ok to me for now. Could you please:
- Bump DESCRIPTION version number to 0.0.2.9000.
- Add an entry to new.md with the corresponding change, as well as in the pkgdown yml file.
- Remove the console.log from the shiny.js JS script.

JohnCoene:
Sorry about the nasty `console.log` it's removed

DivadNojnarg (approved):
LGTM

Issue #417, opened 2024-09-12
Remove console .log


JohnCoene:
There's a `console.log` in here

https://github.com/BristolMyersSquibb/blockr/blame/f08f1066adced31929173b59888045973dab4a56/R/ui.R#L334

Issue #415, opened 2024-09-04
No error reporting


JohnCoene:
This feels a bit funny, I'm not sure what's happening (0 errors found appearing in red)

![image](https://github.com/user-attachments/assets/925e90be-28e7-4e0f-b463-a5a503621452)

DivadNojnarg:
Yes that's a long time standing bug in one of the core validation function, and annoying to see. Thanks for referencing it.

PR #418, opened 2024-09-12
first pass at redesigning submit

DivadNojnarg:
- Improved `submit` feature for blocks. Now submit isn't added as a class but as an optional block attribute. When you design a block, you can pass the `submit` parameter like so:

```r
new_super_block <- function(submit = 0, ...) {
  fields <- list()
  new_block(
    fields = fields,
    expr = quote(print("test")),
    submit = submit,
    ...,
    class = "my_block"
  )
}
```

When `submit = 0`, it will add a submit button in the block toolbar but computations are blocked (clicking on it is required). Internally, each time the `input$submit` is clicked, the submit attribute is incremented by 1 to stay in sync with the UI button. This is useful when the stack is serialised, since this state is kept so that computations can be automatically re-triggered on restore, if `submit > 0`.

```r
serve_stack(new_stack(new_dataset_block(), new_filter_block(columns = "Time", submit = 0)))

# Simulate what happens when restoring a serialised stack
# with submit > 0
serve_stack(new_stack(new_dataset_block(), new_filter_block(columns = "Time", submit = 1)))
```


DivadNojnarg:
Made some tweaks to the button styling. I made it always visible, at the center of the block title, as default.

https://github.com/user-attachments/assets/345db89e-f3ee-47c7-9e19-e60805100d9e

DivadNojnarg:
Button was improved to use the bslib task[^1] button that can't be clicked as long as the task is running.

https://github.com/user-attachments/assets/05db09de-dee8-4e6f-849b-07a76f1e0e83

[^1]: In the future, this can be coupled to `ExtendedTask` if we want to handle async tasks for some blocks and stacks and improve performances).

R/utils.R
```diff
@@ -281,7 +280,7 @@ send_error_to_ui <- function(blk, is_valid, session) {
   )
 
   # Send validation message
-  if (!is_valid$block) {
+  if (!is_valid$block && length(is_valid$message)) {
```
DivadNojnarg, line 283: Fix #415

R/ui.R
```diff
@@ -331,7 +346,6 @@ add_block_ui.default <- function(x, id, ...) {
         tags$head(
           tags$script(HTML("
             function colorText(data) {
-              console.log(data);
```
DivadNojnarg, line 334: Fix #417

DivadNojnarg:
Tweaks to make button more visible. cc @JohnCoene 

Screenshot 2024-09-16 at 12 28 46

JohnCoene:
We'll have to set the version of shiny in `DESCRIPTION` too I believe (since it relies on the latest version)

DivadNojnarg:
@nbenn Feel free to review.

nbenn (commented):
Thanks for looking into this. Some (minor) comments:

- You brought this up yourself in a private msg: should `submit` also be an argument of the "base" ctor? I think *yes* it should. It is a *special* block attribute in the sense that some core behavior depends on these values and adding this to the `new_block()` signature communicates this to a user of our API.
- As a user, I would want to be able to pass a boolean value as `submit`. If I see `submit = 0`, I don't know whether this means `TRUE` or `FALSE` and even if I know `submit = 0` <=> "submit = TRUE", I could then mistakenly assume that `submit = 1` <=> "submit = FALSE". We could allow a flag to be passed as `submit` and then in the base ctor convert `TRUE` -> `0` and `FASLE` -> `-1` or so.
- Personally I'd prefer to have an explicit value as `submit` attribute (e.g. "-1" or so) instead of it not being available corresponding to `submit = FALSE`. This could be resolved just by adding `submit = FALSE` to `new_block()`.
- This is inconsequential, so feel free to ignore: we're only really interested in 3 possible values ("no submit", "submit, but not yet triggered", "already triggered submit"). We could map this to values `c(-1, 0, 1)` or even more compactly `c(FALSE, NA, TRUE)`. Tbh, I'm not super happy w/ my suggestions here. I still feel that differentiating between a value of 543902 and 543903 is pointless. But it also does not hurt.

DivadNojnarg:
@nbenn Included your suggestions and added some unit tests. I am stuck by a shinytest2 issue and Positron. I'd suggest to merge without waiting for this to be solved. I'll open another PR later to add more tests.

nbenn (approved):
Lgtm.

Issue #326, opened 2024-02-29
Shinyline links


JohnCoene:
The `create_app_link` function is an exported function used in the documentation.

https://github.com/blockr-org/blockr/blob/78782f38edd9c45be9ee012dc4c90149e784853b/R/utils.R#L393

How can we easily update the embedded app (embedded by this function)?

DivadNojnarg:
I was working on this https://github.com/DivadNojnarg/shinylivePkgdown2 to automatically export an app to shinylive from the R code. However, I currently don't have availability to finalise it. 

At the moment, the "process" is manual, boring and dumb. You have to copy/paste your app code like:

```r
webr::install("blockr", repos = "https://blockr-org.github.io/webr-repos")
library(blockr)
stack <- new_stack(
    data_block,
    select_block
)

serve_stack(stack)
```

to the Posit shinylive React app editor (https://shinylive.io/r/examples/) and export the encoded url (click on sharing the app button top right corner) after `code=` for instance what I get with the above R code: `NobwRAdghgtgpmAXGKAHVA6ASmANGAYwHsIAXOMpMAdzgCMAnRRASwgGdSoAbbgCgA6YOtyIEA1gyG4ABAzioi7GQF4ZQgBalSqdogD0+kWMkBaIgwDmGSy1IaArnQwsi+2o1PzF7IQEoBCG4WRigGAE8+YwkGAI4uCRkAHlMZCDhqAH1OKAlBCBlCmQATKC5M6PFcQKKZdjhuOAJSCtEJQLjA+oYANzhshPE+HIk-MABfAF0gA` which you pass to `app_code` param. You'll also need a `#nolint` statement.

JohnCoene:
I understand that it just seems a bit complicated to update the documentation, if we want to deploy an updated version of the app we have to change link/code in many places.

DivadNojnarg:
Yes. You also currently have to update the custom wasm repo each time blockr is updated.

That's why this would be much better https://github.com/DivadNojnarg/shinylivePkgdown2  as it would automatically export all apps as soon as you update the code.

DivadNojnarg:
ccing @schloerke :) for another usecase

parmsam-pfizer:
[...]

PR #420, opened 2024-09-16
Update doc

DivadNojnarg:
- Remove `blockr.data` from import. This prevents from installing blockr as the remote field is outdated.
- Update outdated links
- Improve doc according to @MikeJohnPage's feedback
- Improve shinylive links management (cc @JohnCoene) #326
- ...


nbenn (approved):
Lgtm, thanks!

PR #421, opened 2024-09-25
Better result field

nbenn:
Change result field implementation such that
- [x] we no longer need a reactive poll per result field for the data,
- [x] the (serialized) field "value" is only the stack name the field refers to (not data),
- [x] (optional) we no longer need a reactive poll for the stack selector drop-down.


nbenn:
@DivadNojnarg this is a (hacky) start to removing the reactive polls. I think it mostly work (although I am not super confident I did not break any edge-cases).

Previously the "value" of a result field was data. Now it is a string (referring to some stack in the workspace). This comes with a (major) breaking change: while before you could happily do something like

https://github.com/BristolMyersSquibb/blockr/blob/97bea5465a1182946b186d23a5c7ab1dea6c811d/tests/testthat/test-block.R#L154

it no longer makes sense to consider a `result_field` object outside a workspace context. Furthermore, trying to remove the reactive poll for stack names, I "installed" a workspace-level reactive and this now leaves us with something semi-functional in a "result-field-server"-only context like here

https://github.com/BristolMyersSquibb/blockr/blob/97bea5465a1182946b186d23a5c7ab1dea6c811d/tests/testthat/test-result.R#L30-L35

which I worked around with

https://github.com/BristolMyersSquibb/blockr/blob/b18f427df0da06eb9ecec938ca337bb3fce9a57e/R/server.R#L38-L42

(Not sure this is a good idea, but it makes the problem go away.)

I'm currently left with 2 shinytest2-related issues:
- The tests (on gh actions) currently fail due to changes in snapshots. Unfortunately I'm having trouble updating the snapshots due to (unrelated) bootstrap version change issues
    ```
    Diff in snapshot file `blockblock-app-005.json`
    < before
    > after
    @@ 17,5 / 17,5 @@
          "my_stack-block_1-res": {
            "x": {
    <         "style": "bootstrap5",
    >         "style": "bootstrap4",
              "filter": "none",
              "vertical": false,
    @@ 40,41 / 40,5 @@
    ```
    I don't understand why my local environment causes a version downgrade. I did upgrade to the newest bslib, but this did not help. I saw that in some places we explicitly set `bslib::bs_theme(5L)`. Perhaps in other places this could also depend on the session running the app?
- I tried adding my own shinytest2 test for the result field (as the previous `testServer`-approach is kind of pointless with the new implementation). For some reason I cannot get the `AppDriver$new()`-call to produce a "stable" app: I always get `Shiny app did not become stable in x ms`. I tried figuring out whether this comes from some issue I introduce with my change or whether this is a problem with my local shinytest2 setup. Not sure what's going on currently. Do you have any ideas? Can you run the test at

    https://github.com/BristolMyersSquibb/blockr/blob/b18f427df0da06eb9ecec938ca337bb3fce9a57e/tests/testthat/test-result.R#L43-L46

DivadNojnarg:
@nbenn: 

1. To get shinytest2 running, you need to run `devtools::install()` to make sure your tests are running again the latest package version. I initially forgot to do this and got the same error as you.

2. To set the stack names in a running app @JohnCoene made something long time ago. If you click on the stack title, you should be able to edit it. However, this updates only the UI. Is this something we want to bring in this PR too?

JohnCoene:
@DivadNojnarg updating the title is captured I think, it sets the attribute correctly.

DivadNojnarg:
@nbenn and @JohnCoene
- I fixed the test issues.
- Also added some part to account for a title change from the UI. Replaced the `reactive` by `eventReactive` and listening to any stack title attribute change. That was the missing piece causing the result field dropdown not to update properly.

I think, from the UI perspective, it is not obvious that one can click on the title to change it. Maybe we need to add some guidance or a hint about what can be done with the title (like a tooltip "Click to update me...").

DESCRIPTION
```diff
@@ -26,10 +26,10 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.1.9000
 Imports:
     magrittr,
-    shiny,
+    shiny (>= 1.8.1),
     dplyr,
     utils,
-    bslib,
+    bslib (>= 0.7.0),
```
DivadNojnarg, line 32: necessary to get the bslib::input_task_button added in the submit PR. To avoid confusing users.

nbenn:
@DivadNojnarg need your approval, and then we can merge. Tanks for your inputs. Forgot that for shinytest2 we do not have `loal_all()` <=> `install()`.

DivadNojnarg (approved):
Thx!

Issue #423, opened 2024-09-27
Add block


JohnCoene:
In the new app layout we need to dynamically insert the stacks in the UI. This breaks the new add block functionality: no blocks are ever listed.

```r
library(blockr)
library(shiny)

ui <- fluidPage(
  theme = bslib::bs_theme(5L),
  actionButton("create", "Create Stack"),
  div(id = "stacks")
)

server <- function(input, output, session) {
  observeEvent(input$create, {
    stack <- new_stack()

    on.exit({
      generate_server(stack)
    })

    insertUI(
      "#stacks",
      "beforeEnd",
      generate_ui(stack)
    )
  })
}

shinyApp(ui, server
```

@DivadNojnarg would you know what I'm doing wrong?

In previous layouts I could work around that by waiting for the stack UI to be fully rendered before calling the server, however I think it'd be better because this adds a bunch of complexity app-side.

Unless I'm doing something wrong and there's a simple solution :)

DivadNojnarg:
Thanks for checking. I am looking at it.

PR #424, opened 2024-09-27
Add block for dynamically added stack

DivadNojnarg:
Will fix Add block #423 cc @JohnCoene


DivadNojnarg:
@JohnCoene Does this work for you?

JohnCoene:
Just tested, seems to work well, thank you!

PR #430, opened 2024-10-08
hotfix: could not insert result block due to missing check in result field

DivadNojnarg:
@nbenn As discussed, fixing an issue where we could pass NULL to `get_workspace_stack`, leading to an error in `is_string`.

- Bumped version from 0.0.2.9030 to 0.0.2.9031


nbenn (approved):
Lgtm, thanks!

Issue #432, opened 2024-10-08
Result block don't get updated properly


DivadNojnarg:
Our latest PR breaks the result block.

Steps to reproduce:

- Add a stack with dataset block.
- Add a new stack with result block.
- Add select block to first stack and select the first column.

Output: data are not updated in the second stack. This will likely affect the join block too.

nbenn:
@DivadNojnarg 

I have been playing around with this quite a bit and there is one thing that has me confused:

If you start up the app as

```r
serve_workspace(
  stack1 = new_stack(
    new_dataset_block("BOD"),
    new_select_block()
  ),
  stack2 = new_stack(new_result_block("stack1"))
)
```

the app is "broken". If you start up as

```r
serve_workspace(
  stack1 = new_stack(
    new_dataset_block("BOD"),
    new_select_block("Time")
  ),
  stack2 = new_stack(new_result_block("stack1"))
)
```

we're all good.

It seems like the validation failure in the first setting somehow "breaks the reactive chain" and this "breakage" is not cleared up after the issue is fixed.

In either case, I do not understand why `evaluate_block()` is executed in

https://github.com/BristolMyersSquibb/blockr/blob/061ac783f49cb98169e77c845ca851fe0a890354/R/server.R#L225-L232

in the latter case, but not the former case. In both, `blk()` does not change.

Adding to the confusion: If you go ahead and change the data (in "stack1"), you again have an "invalid" state (in the select block) but this time nothing breaks and as soon as the issue is resolved all works fine.

DivadNojnarg:
@nbenn I looked at it a bit more. Could you try from #430 `hotfix2-result-field`? There might still be some edge cases, but it seems to work from your examples.

For the long term, I will continue to work on #427 with the stack validation to hopefully get a more robust solution.

DivadNojnarg:
Summary of findings with @nbenn:

- Putting a break point within the update block observer:
```r
obs$update_blk <- observeEvent(c(r_values(), in_dat(), is_prev_valid()),
        {
          # 1. upd blk,
          if (inherits(blk(), "result_block")) browser()
          b <- update_blk(
            b = blk(),
            value = r_values(),
            is_srv = is_srv,
            input = input,
            data = in_dat()
          )
          blk(b)
          log_debug("Updating block ", class(x)[[1]])

          # 2. Update UI
          update_ui(b = blk(), is_srv = is_srv, session = session, l_init = l_init)
[...]

PR #435, opened 2024-10-14
Attempt to fix Result block don't get updated properly #432

DivadNojnarg:


R/server.R
```diff
@@ -21,6 +21,8 @@ generate_server.result_field <- function(x, ...) {
         res <- get_stack_result(
           get_workspace_stack(inp)
         )
+        # Handle the join block
```
DivadNojnarg, line 24: The join block was broken since we do this:

```r
by_choices <- function(data, y) {
    intersect(colnames(data), colnames(y))
  }
```
 and `y` is a reactive expression coming from the result field, which means it needs to be evaluated to get the correct result. I believe the best place to do this is here, in the server logic rather than in the code block code.

R/server.R
```diff
@@ -702,9 +714,17 @@ generate_server.workspace <- function(x, id, ...) {
       })
 
       attr(x, "reactive_stack_directory") <- reactive({
-        names(vals$stacks)
+        vals$stacks
```
DivadNojnarg, line 720: We need to pass the stacks to do further check within the result field server logic.

R/server.R
```diff
@@ -702,9 +714,17 @@ generate_server.workspace <- function(x, id, ...) {
...
       }) |> bindEvent(
-        chr_ply(lapply(vals$stacks, `[[`, "stack"), attr, "title")
+        c(
+          chr_ply(lapply(vals$stacks, `[[`, "stack"), attr, "title"),
+          lapply(vals$stack, \(stack) {
```
DivadNojnarg, line 724: added an extra trigger each time a block changes and it's validity change, otherwise the result field does not update consistently.

R/server.R
```diff
@@ -396,10 +403,15 @@ generate_server.stack <- function(x, id = NULL, new_block = NULL,
       # Any block change: data or input should be sent
       # up to the stack so we can properly serialise.
       observeEvent(
-        c(
-          get_block_vals(vals$blocks),
-          get_last_block_data(vals$blocks)()
-        ),
+        {
+          lapply(vals$blocks, \(block) {
+            req(block)
```
DivadNojnarg, line 411: a little extra check to prevent this event from triggering when vals$block is a list of NULL elements.

R/server.R
```diff
@@ -44,13 +44,18 @@ generate_server.result_field <- function(x, ...) {
         updateSelectInput(
           session,
           "select-stack",
-          choices = result_field_stack_opts(session$ns, workspace_stacks()),
+          choices = result_field_stack_opts(session$ns, names(workspace_stacks())),
```
DivadNojnarg, line 49: Since we pass the entire stack object in the `reactive_stack_directory`, we just extract the names to display them in the select choices.

R/server.R
```diff
@@ -210,7 +219,7 @@ generate_server_block <- function(
       out_dat <- if (attr(x, "submit") > -1) {
         eventReactive(input$submit,
           {
-            req(is_valid$block)
```
DivadNojnarg, line 213: req was blocking the reactive chain. We just return an empty data.frame if invalid.

R/server.R
```diff
@@ -223,10 +232,11 @@ generate_server_block <- function(
         )
       } else {
         reactive({
-          req(is_valid$block)
+          if (!is_valid$block) return(data.frame())
           if (is.null(in_dat()) && !inherits(x, "transform_block")) {
             evaluate_block(blk())
           } else {
+            if (nrow(in_dat()) == 0 && !inherits(x, "parser_block")) return(data.frame())
```
DivadNojnarg, line 239: `parser_block` are an exception as their input isn't rectangular data but a file path.

R/server.R
```diff
@@ -44,13 +44,18 @@ generate_server.result_field <- function(x, ...) {
...
       )
 
       reactive({
-        get_result(input[["select-stack"]])
+        stacks <- workspace_stacks()
+        blocks <- stacks[[input[["select-stack"]]]]$blocks
+        all_valid <- lgl_ply(blocks, \(block) {
+          block$is_valid()
+        })
+        if (all(all_valid) == TRUE) get_result(input[["select-stack"]]) else data.frame()
```
DivadNojnarg, line 61: Overall, we check that the previous stack is valid (otherwise the result block has no way to know about the previous blocks from the linked stack). Note: this will be superseded in the stack validation PR #427.

R/server.R
DivadNojnarg: I actually replaced this by:

```r
if (inherits(in_dat(), "data.frame") && nrow(in_dat()) == 0) {
              return(data.frame())
            }
```

as it broke the plots.

R/server.R
```diff
@@ -236,7 +236,9 @@ generate_server_block <- function(
           if (is.null(in_dat()) && !inherits(x, "transform_block")) {
             evaluate_block(blk())
           } else {
-            if (nrow(in_dat()) == 0 && !inherits(x, "parser_block")) return(data.frame())
```
DivadNojnarg: sorry this thing was a dumb idea ...

R/server.R
nbenn: This shouldn't cause problems: `y` from "within the block" should not behave as reactive, but as regular R value. All of this should happen in `value()`:

https://github.com/BristolMyersSquibb/blockr/blob/061ac783f49cb98169e77c845ca851fe0a890354/R/field-core.R#L184-L209

So if this is still an open issue (is it?), the question is why the "reactive representation" slips through here.

R/server.R
nbenn: Just for my understanding: why do we need to pass through the entire stacks here? Initially the intention for the `reactive_stack_directory` was just to give a reactive object for populating result field drop-downs. Where do we need the reactive stack? Sorry, I didn't spot anything relevant in your changes.

nbenn: I see, it is the validity check for the dependent stack that this is used for. If this is the chase, the result field validator should take care of this, no? Can we extend

https://github.com/BristolMyersSquibb/blockr/blob/061ac783f49cb98169e77c845ca851fe0a890354/R/fields.R#L298-L312

to not only check for existence of a stack, but also it's "validity"?

DivadNojnarg: I played around this idea in https://github.com/BristolMyersSquibb/blockr/commit/5822c89bdf0b4253cf91f8ec7124ea7141f92721. However, while it works well for a "static" case, that is, with predefined stacks, it fails in the dynamic case when block are added on the fly:

```r
# works
serve_workspace(
  stack1 = new_stack(new_dataset_block("BOD"), new_select_block()),
  stack2 = new_stack(new_result_block("stack1"))
)

# fails: to replicate, add a new_select_block on the fly and nothing happens.
serve_workspace(
  stack1 = new_stack(new_dataset_block("BOD")),
  stack2 = new_stack(new_result_block("stack1"))
)
```

R/server.R
DivadNojnarg: Sorry. missed this part of the code. Debugging shows we never enter the first if statement as `list_workspace_stacks()` always return character(0).

R/server.R
DivadNojnarg: Actually this just does not work. My workspace was already contaminated by other existing stacks so I though it worked but no ...

PR #439, opened 2024-10-21
Minimally invasive result field fix

nbenn:
For my two example cases, this works. Have not checked more "exotic" scenarios such as plot blocks etc.

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -2,7 +2,8 @@ linters: linters_with_defaults(
   line_length_linter = line_length_linter(100L),
   object_name_linter = NULL, # Because we use S3 and end up with is_initialized.field
   object_usage_linter = NULL,   # When code is WIP this is annoying ...
-  commented_code_linter = NULL  # When code is WIP this is annoying ...
+  commented_code_linter = NULL,  # When code is WIP this is annoying ...
+  complexity_limit = 16L
   )
 exclusions: list(
     "inst/examples/cdisc-plot/example.R",
--- a/.lintr
+++ b/.lintr
@@ -3,7 +3,7 @@ linters: linters_with_defaults(
   object_name_linter = NULL, # Because we use S3 and end up with is_initialized.field
   object_usage_linter = NULL,   # When code is WIP this is annoying ...
   commented_code_linter = NULL,  # When code is WIP this is annoying ...
-  complexity_limit = 16L
+  cyclocomp_linter(16L)
   )
 exclusions: list(
     "inst/examples/cdisc-plot/example.R",
--- a/R/block-core.R
+++ b/R/block-core.R
@@ -220,13 +220,18 @@ generate_code.call <- function(x) {
 #' @inherit new_block
 #' @param ... For generic consistency.
 #' @export
-evaluate_block <- function(x, ...) {
+evaluate_block <- function(x, data, ...) {
+
+  if (!missing(data) && is.null(data)) {
+    return(NULL)
+  }
+
   UseMethod("evaluate_block")
 }
 
 #' @rdname evaluate_block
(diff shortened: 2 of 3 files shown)


nbenn:
@DivadNojnarg many thanks for the pointer

> if you remove req(is_valid$block) and replace by if (!is_valid$block) return(data.frame())

DivadNojnarg:
@nbenn  you probably want to remove the `expect_error` to fix the broken test: https://github.com/BristolMyersSquibb/blockr/pull/435/files#diff-e3fcd79bc10ab1ead4020ff2d87bb6d914b30e88828a04babc7b51abf8ccea5a as we don't use `req` anymore.

nbenn:
Good points @DivadNojnarg, thanks! I was suspecting that there might be a bit of fallout with this `NULL` result value. But I do feel that moving in such a direction is probably good (to have a generic `NULL` instead of more "block-specific" values such as `data.frame()` etc.). Do you agree with this sentiment?

(Maybe the "minimally-invasive" part of the title was a bit misleading..)

DivadNojnarg:
> Do you agree with this sentiment?

Yes I prefer NULL over having to specify different empty constructors.

nbenn:
@DivadNojnarg do you know why we do

https://github.com/BristolMyersSquibb/blockr/blob/061ac783f49cb98169e77c845ca851fe0a890354/R/server.R#L227

in the first place? Wouldn't this better be something like `if (inherits(x, "data_block")`?

(Side remark: the logic in

https://github.com/BristolMyersSquibb/blockr/blob/061ac783f49cb98169e77c845ca851fe0a890354/R/server.R#L214

should be exactly the same, no?)

For the other issue you point out:

> For instance, at the moment, if you empty the data selection in the first block, the app crashes since we try to select on an empty data.frame

High-level, I feel the problem is somewhere else: if you empty the dataset selection, the datase block should turn "invalid" and the select block should then not evaluate. So we'd need to factor in `is_prev_valid` here too. Does that sound reasonable?

nbenn:
@DivadNojnarg 

> We need some safe gards whenever the data passed from one block to another is NULL and the block isn't a data block.

Does this happen, i.e. that we have `NULL` as a result of a "valid" block? I was hoping that this could be avoided.

DivadNojnarg:
The first point is likely an accidental removal. They should have the same conditions.

Second point: something like:

```r
reactive({
          if (!is_valid$block || (!is.null(is_prev_valid()) && !isTRUE(is_prev_valid()))) {
            NULL
          } else if (is.null(in_dat()) && !inherits(x, "transform_block") && !inherits(x, "plot_block")) {
            evaluate_block(blk())
          } else {
            evaluate_block(blk(), data = in_dat())
          }
        })
```

DivadNojnarg:
In the update block observer, we also may want to prevent to update the block if the previous one is invalid:

```r
obs$update_blk <- observeEvent(c(r_values(), in_dat(), is_prev_valid()),
        {
          if (!is.null(is_prev_valid()) && !isTRUE(is_prev_valid())) return(NULL)
         ...
})
```

An even better alternative would be to avoid triggering the event at all. However, with our current logic, you can't just do req(is_prev_valid()) since for data block is_prev_valid() is always NULL, and this would block the whole chain. That said, it could be rewritten a bit like so:

```r
# I must admit this thing should be done with different S3 server methods, but as we don't have them anymore ...
update_blk_trigger <- reactive({
   if (inherits(x, "data_block")) {
     r_values()
   } else {
     c(r_values(), in_dat(), req(is_prev_valid()))
   }
})

obs$update_blk <- observeEvent(update_blk_trigger(), { ...})
```

Issue #467, opened 2026-03-13
dependencies on dev versions are not strict enough due to breaking changes


moodymudskipper:
I had this:

```
library(blockr)                                                                                                                                                            
  +                                                                                                                                                                                          
  + run_app(                                                                                                                                                                                     
  +   blocks = c(                                                                                                                                                                              
  +     a = new_dataset_block("iris")                                                                                                                                                          
  +   )                                                                                                                                                                                      
  + )                                                                                                                                                                                        
  ── Attaching blockr packages ─────────────────────────────────────────────────────────────────────────────────────────────────────── blockr 0.1.1 ──                                       
  ✔ blockr.core   0.1.2     ✔ blockr.dplyr  0.1.0                                                                                                                                            
  ✔ blockr.dag    0.1.0     ✔ blockr.ggplot 0.1.0                                                                                                                                            
  ✔ blockr.dock   0.1.0     ✔ blockr.io     0.1.0                                                                                                                                            
  Warning messages:                                                                                                                                                                          
  1: package ‘blockr.dock’ was built under R version 4.4.3                                                                                                                                   
  2: package ‘blockr.dag’ was built under R version 4.4.3                                                                                                                                    
  3: package ‘blockr.dplyr’ was built under R version 4.4.3                                                                                                                                  
  4: package ‘blockr.ggplot’ was built under R version 4.4.3                                                                                                                                 
  5: package ‘blockr.io’ was built under R version 4.4.3                                                                                                                                     
  Warning: Error in UseMethod: no applicable method for 'as_blocks' applied to an object of class "character"                                                                                
    88: as_blocks                                                                                                                                                                            
    84: block_metadata                                                                                                                                                                       
    83: blks_metadata                                                                                                                                                                        
    82: edit_ui                                                                                                                                                                              
    79: <Anonymous>                                                                                                                                                                          
    78: mapply                                                                                                                                                                               
    77: Map                                                                                                                                                                                  
    76: map                                                                                                                                                                                  
    75: block_ui.dock_board                                                                                                                                                                  
    66: board_ui.dock_board                                                                                                                                                                  
    63: blockr_app_ui.dock_board                                                                                                                                                             
    60: ui                                                                                                                                                                                   
     3: runApp                                                                                                                                                                               
     2: print.shiny.appobj                                                                                                                                                                   
     1: <Anonymous>                                                                                                                                                                          
                                                                                                                                                                                             
                                                                                                                                                                                             
                                                                                                                                                                                             
[...]

PR #468, opened 2026-03-13
Fix missing version constraints and GitHub remotes for dependencies

moodymudskipper:
For #467, though maybe not robust enough, other dependencies might have minimal versions to adjust

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,8 +47,7 @@ Imports:
     rlang
 Remotes:
     BristolMyersSquibb/blockr.dock,
-    BristolMyersSquibb/blockr.dag,
-    cynkra/g6R
+    BristolMyersSquibb/blockr.dag
 Suggests:
     testthat (>= 3.0.0),
     quarto,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,15 +36,19 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.3
 Config/testthat/edition: 3
 Imports:
-    blockr.core,
-    blockr.dock,
-    blockr.dag,
+    blockr.core (>= 0.1.2),
+    blockr.dock (>= 0.1.1),
+    blockr.dag (>= 0.1.2),
     blockr.dplyr,
     blockr.ggplot,
     blockr.io,
     utils,
     cli,
     rlang
+Remotes:
+    BristolMyersSquibb/blockr.dock,
+    BristolMyersSquibb/blockr.dag,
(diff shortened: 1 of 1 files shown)


nbenn:
There was an attempt to fix the issue of having to add/rm remotes for CRAN releases and for not constantly having to change remotes specs for PRs to get CI to pass. It's currently a bit half-assed and not well tested tbh. See the README of [cynkra/blockr.ci](https://github.com/cynkra/blockr.ci) for some more details.

The main thing to know is that you can in a PR body require a certain ref to be used for some of the pkg deps. There's also some logic that [tracks "internal" packages](https://github.com/cynkra/blockr.ci/blob/main/.github/actions/registry.txt).

To wrap that into usable "dev install logic", we have something like

```r
source("https://raw.githubusercontent.com/cynkra/blockr.ci/main/inst/install.R")
install_blockr(Sys.getenv("REPO"))
```

which then for each repo is available as pages artifact e.g. for `blockr.core` you'd run

```r
source("https://bristolmyerssquibb.github.io/blockr.core/install")
```

and get all dev deps. But again, it's half-baked atm. We'd nee to update some READMEs for sure. And building on the "not well tested" comment from above, i'd not be surprised if there were not also some wrinkles to iron out still. Also, this repo for one has not been updated to use the latest (and centralized) CI setup.

A more general comment: we decided against including Claude attributions (Cc @christophsax). Something like


```json
{
  "attribution": {
    "commit": "",
    "pr": ""
  }
}
```

in your Claude `settings.json` should do.

 (Cc @DivadNojnarg, given this discussion spans multiple repos/issues.)

nbenn:
Thinking about this some more, I feel the approach as we have it atm is a bit off. This non-canonical install routine causes too much friction. So I think we should add back proper dependency specification and have a claude skill or so that helps with friction caused by the update commits needed pre/post merges.

What we still need in order to get clean check results (for the gh-based revdep checks) is a way to in-PR specify against what ref the revdep checks should be run.

@DivadNojnarg thoughts? Does that sound reasonable?

moodymudskipper:
If we enforce always bumping the package version when pushing to main, using fledge maybe, we can have all other packages depending on the latest version of their dependencies using remote at all times (all newest main depend on other newest main), it can be automated through CI.

That would at least make it possible for users and developers to use development versions reliably. 

Ideally though every new dev package dependency should be compatible with old CRAN dependents (with fallbacks to old code if necessary), or the CRAN update process will be chaotic. New dependents can require new versions of dependencies in the DESCRIPTION so in this direction we can just require latest always.

PR #482, opened 2026-06-17
Forward plugins through run_app() to serve()

nbenn:
## Summary

- `run_app()` passed only `id` on to `serve()`, swallowing `plugins` and `options` into `...` (where they reached `new_dock_board()` instead). Both are now explicit formals forwarded to `serve()`, each defaulting to serve's own `blockr_app_plugins` / `blockr_app_options`, so no-argument behaviour is unchanged.
- This exposes serve's customization pair through the meta-package entry point: `run_app(plugins = custom_plugins(…), options = custom_options(…))`. Board-option *values* stay reachable via `custom_options()` — the serve-level parallel to `custom_plugins()` — so routing `options` to `serve()` rather than `new_dock_board()` loses nothing; only the raw `run_app(options = <board_options>)` constructor path changes, which is fine in a forward-only ecosystem.
- The `options` formal is referenced as `base::options()` in the body, so the option-setting call there still hits the base function rather than the new formal.

Companion to BristolMyersSquibb/blockr.uat#2, whose app needs `run_app(plugins = custom_plugins(manage_project()))`. That repo pins `blockr` to this branch until this merges, then flips the pin to main.

### DIFF
--- a/R/run.R
+++ b/R/run.R
@@ -20,13 +20,16 @@
 #' @rdname run_app
 #' @export
 run_app <- function(..., extensions = new_dag_extension(),
-                    id = rand_names()) {
+                    plugins = blockr_app_plugins,
+                    options = blockr_app_options, id = rand_names()) {
 
-  prev <- options(g6R.layout_on_data_change = TRUE)
-  on.exit(do.call(options, prev))
+  prev <- base::options(g6R.layout_on_data_change = TRUE)
+  on.exit(do.call(base::options, prev))
 
   serve(
     new_dock_board(..., extensions = extensions),
-    id = id
+    id = id,
+    plugins = plugins,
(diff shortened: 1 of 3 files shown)


nbenn:
Superseded by #484, which landed the same `run_app()` plugins forwarding (merged in df486ca). #484 takes the cleaner route: `plugins` is forwarded to `serve()` while `options` stays routed to `new_dock_board()` as board-option values — which serve's default resolver reads back via `board_options()` — so the `base::options()` shadow and the forced `custom_options()` wrapper both go away and `run_app(options = <values>)` keeps working. Leaving the `run-app-plugins` branch in place for now since blockr.uat#3 still pins it; that pin flips to `blockr@main` as the follow-up.

